PR #2, opened 2016-03-04
add_benchmark should be immune to absolute paths

brunocodutra:
this way one may safely 

`add_benchmark(path.to.benchmark "${CMAKE_CURRENT_SOURCE_DIR}/path/to/benchmark")`

### DIFF
--- a/metabench.cmake
+++ b/metabench.cmake
@@ -37,11 +37,14 @@ endif()
 # prior to being evaluated as an ERB template. This can be used to include
 # platform-dependent informations in the `chart.json` file.
 function(add_benchmark target path_to_dir)
+    string(REPLACE "${CMAKE_CURRENT_SOURCE_DIR}/" "" path_to_dir ${path_to_dir})
+
     # Dependencies of the benchmark; the benchmark will be considered
     # outdated when any of these is changed.
-    file(GLOB dependencies ${path_to_dir}/chart.json
-                           ${path_to_dir}/*.cpp
-                           ${path_to_dir}/*.hpp)
+    file(GLOB dependencies
+        "${CMAKE_CURRENT_SOURCE_DIR}/${path_to_dir}/chart.json"
+        "${CMAKE_CURRENT_SOURCE_DIR}/${path_to_dir}/*.cpp"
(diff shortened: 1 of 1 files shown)


brunocodutra:
Because otherwise if an absolute path is used you'd have `_metabench/a/hell/of/a/long/prefix/path/to/benchmark` instead of the expected `_metabench/path/to/benchmark`.

ldionne:
Well no, because you would just have made the path relative with `string(REPLACE ...)`?

ldionne:
In other words, we're making the `path_to_dir` relative to the current source directory using the `string(REPLACE ...)`. From then on, it should be safe to assume that the path is relative to the current source dir, and so we can use the relative path in the `file(GLOB ...)` call.

brunocodutra:
I don't get it, what you mean?
The idea is to strip `${CMAKE_CURRENT_SOURCE_DIR}` from `path_to_dir`, since the latter is used to create the output json file.
If we don't erase `${CMAKE_CURRENT_SOURCE_DIR}`, the we would get a very long prefix.

And I'm pretty sure CMake deals with such windows nastiness for us.

brunocodutra:
To test, try this out `add_benchmark(path.to.benchmark "${CMAKE_CURRENT_SOURCE_DIR}/path/to/benchmark")`

ldionne:
I understand the purpose of line 40 of your patch, but not lines 44 to 47. That is my question.

brunocodutra:
Oh I see, I don't know, they are probably unnecessary, but I taught myself to be strict when it comes to pathnames because I've shot myself in the foot far too many times due to stupid relative paths.

I can revert those lines if you want.

brunocodutra:
I mean, they don't hurt either right?

ldionne:
Ok, I just wanted to see whether I was missing something. It's fine this way.

brunocodutra:
Oh just realized you can actually comment pull requests line by line.
Sorry for the mess.

ldionne:
> Oh just realized you can actually comment pull requests line by line.

Yes, it's handy! No problem!

PR #6, opened 2016-03-07
metabench_add_benchmark should use current CMAKE_CXX_COMPILER and CMAKE_CXX_FLAGS

brunocodutra:
This avoids the redundant cxx parameter in the chart.json and also allows common compile flags to be passed onto various benchmarks instead of specifying them per benchmark via chart.json

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -7,4 +7,5 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/..")
 
 include(metabench)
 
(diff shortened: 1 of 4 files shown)


brunocodutra:
it is beginning in my previous commit

ldionne:
Yeah, I just realized this. Sorry for the noise. P.S.: You can reply to my inline comment by pressing "add a line note" [here](https://github.com/brunocodutra/metabench/commit/a631a876a93ec946611c0dbfab75483f34672461).

ldionne:
Merged in 74d5fdf18c19f75f36d2abbeed26525ec3e4799f and 76578fe107b486a0e5662c2dc365831e67271e11, but GitHub is not clever enough to see it.

PR #7, opened 2016-03-07
the download of highcharts.js shouldn't fail silently

brunocodutra:
if the download fails we could either:

1. error out
2. emit some warning and do not generate the html file
3. fall back to loading highcharts at the time of visualization
4. fall back to some hardcoded version of highcharts

I implemented number 3 so at least we try providing the visualization, but perhaps we should go for number 4 to make it completely independent of internet connection?


ldionne:
Solutions (1)  and (2) do not seem optimal, although we could mix (2) and (3) and emit a warning saying what we did. Solution (4) would be the best, but it would be very tedious to hardcode highcharts into the CMake module (we'd have to escape a bunch of stuff). And I want to avoid shipping more than a single CMake file if possible, so shipping highcharts.js in a separate file is not an option.

ldionne:
0adc05462a9cd7b98c441f02e4caa862d2587dbf adds a warning message when we fail to download Highcharts.

brunocodutra:
Regarding option 4, wouldn't it be just a matter of escaping double quotes and at signs? 
My main concern regarding this option are actually legal issues.

ldionne:
They say it's free for non-commercial purposes. I think we could get by by stating which part of Metabench's code is under their license (the part that would be hardcoded, basically).

However, the escaping of everything properly is my concern. How would you do that?

brunocodutra:
In fact I'm rather concerned about the use of highcharts, as its "non-commercial use" license seems to imply metabench couldn't be integrated to commercial projects.
As there are plenty other equally powerful charting tools, I'd vote for migrating away from highcharts. I'd suggest [nvd3](https://github.com/novus/nvd3) which is licensed under Apache 2.0.

Ragarding escaping, wouldn't a simple regex-less find&replace do the trick?

ldionne:
Honestly, I don't know how much this is a problem. I mean, it seems unlikely that someone needs to use Metabench for some commercial purpose, no? The main problem I see is a technical one: it's already written using Highcharts and it works. Plus, Hana contains a decent amount of charts written using Highcharts, so that would be slightly a pain to change.

I'm just trying to evaluate the return on investment here, and I'm not sure changing the charting library is a wise move.

brunocodutra:
Well I'm really not sure of the implications of such licenses as Highchart's, I mean, the fact that you use it to generate plots from within Hana wouldn't mean Hana as a whole couldn't be used commercially? I really don't know and would risk it myself. 

I do understand the implications of moving away from Highcharts for Hana in particular. Perhaps it would be a good idea to at least contact them directly for advice.

ldionne:
If I was using Highcharts in Hana's source code, then I would need to publish Hana under a dual license (Boost + Highcharts'), and that would mean that the source code can only be used in non-commercial applications (per the most restrictive license, which would be Highcharts'). But this is not the case, and the charts are only there "in support" to the rest of the project. That's my understanding, but I'll contact them to make sure.

brunocodutra:
It does make sense. 

I think it would be nice if you could post their answer here for reference.

ldionne:
Ah crap, I just read this in their FAQ:

> __May I use your Software under the Non-Commercial License for Open Source Projects?__

> Although Highsoft´s Software have open source codes, our software is not licensed as an open source software and are unfortunately not compatible with any open source software license like Apache 2 or any GPL. See also Non-Commercial Redistribution

ldionne:
Let's continue this in #9.

PR #8, opened 2016-03-07
Add support for Appveyor

### LINKED ISSUE #4: Metabench should  be tested on windows
I think it wouldn't be too hard to set up Appveyor.

ldionne:
This pull request adds support for testing Metabench on Windows. Once merged, this will fix #4.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,59 @@
+# Copyright Louis Dionne 2016
+# Copyright Bruno Dutra 2016
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+
+shallow_clone: true
+
+branches:
+  except:
+    - /pr\/.+/
+
+os:
+  - Visual Studio 2015
+
+build:
+  verbosity: detailed
+
+configuration:
+  - Debug
+
+install:
+  - choco install ninja
+
+  ############################################################################
+  # All external dependencies are installed in C:\projects\deps
+  ############################################################################
+  - mkdir C:\projects\deps
+  - cd C:\projects\deps
+
+  ############################################################################
+  # Install a pre-built Clang-cl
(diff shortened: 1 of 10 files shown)


brunocodutra:
just a heads up, instead of setting `-std=c++11`, which is not understood by MSVC, we can raise CMake minimum version to 3.1 and use the following instead

`set(CMAKE_CXX_STANDARD 11)`
`set(CMAKE_CXX_STANDARD_REQUIRED ON)`

This way we could test MSVC on windows, since we are delegating the compilation to CMake already anyways.

ldionne:
Yup, thanks, I'm already doing this :) But I didn't know it was CMake >= 3.1, so I'll update that. I'm almost there for Appveyor support.

brunocodutra:
crap, just realized it :I

but yeah, I think it is >= 3.1.

brunocodutra:
So is there anything still preventing us from testing MSVC?

ldionne:
Technically, no, although I wanted to first get it working with Clang-cl because I already had the Appveyor script from Hana and I've no idea how to use MSVC.

brunocodutra:
[I use it to test Metal](https://github.com/brunocodutra/metal/blob/master/appveyor.yml), but I don't think I'm using a "Makefile Generator" there.

ldionne:
No, you're not using a Makefile Generator. It's certainly possible to pass MSVC as `CMAKE_CXX_COMPILER`, though?

brunocodutra:
I'd guess we could use "NMake Makefiles" and "NMake Makefiles JOM", while setting CMAKE_CXX_COMPILER to `<path to msvc binaries>/cl.exe`

brunocodutra:
Perhaps when calling `vcvarsall.bat` the appropriate `cl.exe` is added to PATH, so we wouldn't need the absolute path.

ldionne:
Now that basic Appveyor support is locked down (finally...), let's move to #18.

PR #10, opened 2016-03-08
target dependencies shouldn't be filtered by extensions

brunocodutra:
rationale: a benchmark is a directory and should depend on anything within


ldionne:
Makes sense.

Issue #9, opened 2016-03-07
Move away from Highcharts


ldionne:
Highcharts' license is not compatible with open source licenses, so we'll have to move away from them.

brunocodutra:
It seems like [Chart.js](https://github.com/nnnick/Chart.js) and [nvd3](https://github.com/novus/nvd3) would be good alternatives.

brunocodutra:
Closing in favor of #13.

brunocodutra:
This should be closed in favor of #13 .

PR #13, opened 2016-03-09
Move away from Highcharts

brunocodutra:
This pull request addresses #9


brunocodutra:
metabench.cmake is now REALLY cluttered by thousands of lines of escaped javascript code, but I think the benefit of being totally offline is worth it.

ldionne:
Looks cool. I assume you used a script to escape the javascript? Perhaps it would be a good idea to add the script to the repository if we ever want to update the dependencies.

ldionne:
Also, do you know how we could provide a title (and subtitle) to the chart? I searched on Google, but it seems like we can't do it from the JSON and we'd have to do it in JavaScript.

brunocodutra:
Nothing fancy, just very simple and hacky perl scripts to insert a line break every 80 characters, then escape backslashes and double quotes. I'll add it as a self-contained script to `tools/` later today.

NVD3 doesn't seem have a builtin way to add titles to `lineChart`s, but we could always do it manually using javascript. Indeed, if you look closer you will realize that I actually manually extract information from the JSON in order to setup the chart, as there's no `renderto` equivalent in NVD3. I'll add something to dynamically create labels based on two new properties `title` and `subtitle`. Would it be enough to just expect the string to be displayed?

ldionne:
> Would it be enough to just expect the string to be displayed?

Yes, it would be enough.

brunocodutra:
This should be ready for merging now.

ldionne:
Awesome, thanks a lot!

PR #14, opened 2016-03-09
Run the benchmarks through CMake

ldionne:
This experimental pull request runs benchmarks through CMake instead of running them "by hand". The advantages are

1. In theory, the benchmarks should run on any platform since we're using CMake to run the benchmarks
2. We do not need to go through hoops to get `CXXFLAGS` and similar properties to the benchmark. Everything defined in CMake will automatically be used by the benchmarks.

On the other hand, this has the following downsides:

1. This causes several "dummy" targets to be created, thus polluting the list of available target when calling e.g. `ninja help`. For each benchmark, this is going to generate a `_metabench.target` target, which is used by Metabench to run the benchmark.
2. The benchmarks are slower to run, because we use CMake to run them every time and must sometimes wait for the dummy targets to get out of date. It might be possible to work around this, but basically it means that for now the benchmarks will take longer to complete.

FWIW, this approach is what Hana currently does. It works well, but as I outlined above there are also disadvantages. I'm putting this PR here to gather comments.


brunocodutra:
Excelent!

ldionne:
So you think the portability is worth the added complexity and downsides?

metabench.cmake
```diff
@@ -70,6 +70,14 @@ function(metabench_add_benchmark target path_to_dir)
         message(FATAL_ERROR "Path specified to add_benchmark (${path_to_dir}) does not contain a chart.json file.")
     endif()
 
+    # Create a dummy executable that will be used to run the benchmark. We'll
+    # run the benchmark through CMake, which allows us to be cross-platform
+    # without additional work.
+    file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/_metabench/${path_to_dir}/measure.cpp "")
+    add_executable(_metabench.${target} EXCLUDE_FROM_ALL ${CMAKE_CURRENT_BINARY_DIR}/_metabench/${path_to_dir}/measure.cpp)
+    set_target_properties(_metabench.${target} PROPERTIES RULE_LAUNCH_COMPILE "${RUBY_EXECUTABLE} -- ${MEASURE_RB_PATH}")
+    set_property(TARGET _metabench.${target} APPEND PROPERTY INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}/${path_to_dir}")
```
brunocodutra: is this equivalent to the less verbose [target_include_directories](https://cmake.org/cmake/help/v3.0/command/target_include_directories.html)?

metabench.cmake
```diff
@@ -70,6 +70,14 @@ function(metabench_add_benchmark target path_to_dir)
         message(FATAL_ERROR "Path specified to add_benchmark (${path_to_dir}) does not contain a chart.json file.")
     endif()
 
+    # Create a dummy executable that will be used to run the benchmark. We'll
+    # run the benchmark through CMake, which allows us to be cross-platform
+    # without additional work.
+    file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/_metabench/${path_to_dir}/measure.cpp "")
+    add_executable(_metabench.${target} EXCLUDE_FROM_ALL ${CMAKE_CURRENT_BINARY_DIR}/_metabench/${path_to_dir}/measure.cpp)
+    set_target_properties(_metabench.${target} PROPERTIES RULE_LAUNCH_COMPILE "${RUBY_EXECUTABLE} -- ${MEASURE_RB_PATH}")
```
brunocodutra: it appears ["non-Makefile generators currently ignore this property"](https://cmake.org/cmake/help/v3.0/prop_gbl/RULE_LAUNCH_COMPILE.html).

I'm not sure what a "non-Makefile generator" is, but does that mean MSVC generators aren't supported?

metabench.cmake
```diff
@@ -152,24 +161,22 @@ file(WRITE ${METABENCH_RB_PATH}
...
+"      stdout, stderr, status = Open3.capture3(command)                                            \n"
+"      command_line = stdout.match(/\\[command line: (.+)\\]/i)                                    \n"
 "                                                                                                  \n"
-"            error_string = <<-EOS                                                                 \n"
-"compilation error: #{command}                                                                     \n"
+"      # If we didn't match anything, that's because we went too fast, CMake                       \n"
+"      # did not have the time to see the changes to the measure file and                          \n"
+"      # the target was not rebuilt. So we sleep for a bit and then retry                          \n"
+"      # this iteration.                                                                           \n"
+"      (sleep 0.2; redo) if command_line.nil?                                                      \n"
```
brunocodutra: Isn't there a way to have Ruby monitor a file for changes and only trigger the compilation uppon this event?

The idea is to emulate what CMake does to decide whether the target is up to date.

metabench.cmake
ldionne: I think so. Fixed.

metabench.cmake
ldionne: Yes, I assume this is what it means. The Makefile generators are, I assume, those listed as "Command-Line Build Tool Generators" [here](https://cmake.org/cmake/help/v3.0/manual/cmake-generators.7.html#cmake-generators). Is the fact that IDE generators don't work a dealbreaker?

metabench.cmake
ldionne: The problem is that when we run the target through CMake, CMake will check whether the file is outdated too, and it will determine that it isn't. I think it has to do with the timestamps on the file not being granular enough, which is not a problem with CMake per se.

ldionne: Ah, I think I understand your comment. You want us to monitor the file and then call CMake as soon as the filesystem has determined that the file changed? So basically we'd be waiting for `min(time for filesystem to update the timestamp, 0.2)`, which would always be less than `0.2`. Right?

metabench.cmake
brunocodutra: Not for me, looks good.

metabench.cmake
brunocodutra: Yeah something like that, we basically try to emulate what CMake does itself to determine whether the target is outdated (not sure what that would be though).

I think you are considering doing it by polling the filesystem every 0.2 seconds. Forgive my ignorance, but wouldn't there be an easy way to asynchronously wait for some event to be triggered?

brunocodutra:
> So you think the portability is worth the added complexity and downsides?

Definitelly, I find it acutally rather an elegant solution.

brunocodutra:
It would be cool if the user were able to manually setup the benchmark target just like one does to executable/library targets through `set_target_properties` for instance.

I'm quite sure there's a way to make targets inherit properties, though I've never used it myself. Perhaps we could have `_metabench.${target}` inherit `${target}`'s properties? This way the user could add cxx flags per benchmark target besides setting CMAKE_CXX_FLAGS prior to adding a benchmark.

ldionne:
Yes, it would be neat, but I'm not sure how to do this, or if it's even possible. I'll start by polishing this out and merging it, then we can improve on both the waiting time and the way to set target properties.

brunocodutra:
All right.

PR #19, opened 2016-03-10
Add the .erb suffix to files inside a benchmark directory

ldionne:
This is a simple commit, but I'm putting it here to request comments.

The argument for is that it's more consistent, because these files are actually ERB files. It can also help text editors for highlighting purposes. On the other hand, text editors will _not_ highlight C++ code in the `.cpp.erb` files, which can be annoying. So this is a two-edged sword.


brunocodutra:
That's exactly why I removed the extension filter, I intend to add `.erb` to benchmarks in Metal.

ldionne:
Alright, so that answers the question.

Issue #18, opened 2016-03-10
Add testing on MSVC


ldionne:
Right now, Appveyor only tests on Clang-cl. We should also test on MSVC.

ldionne:
Preliminary trials suggest that this is just a matter of setting `CMAKE_CXX_COMPILER=cl` instead of `clang-cl`.

brunocodutra:
Would it make any difference to use other versions of MSVC, I mean, does it depend on the compiler at all or just on the generator?

Also, shouldn't we explicitly test NMake and NMake JOM to make sure it works with both of them?

ldionne:
I'm pretty sure it should work with any version of MSVC, provided they can compile the example benchmarks (which need C++11). And yes, we should probably test the other generators to make sure they work, but I've no experience with them. Have you used these NMake generators?

brunocodutra:
Not like this no, but leave this on me, I'll figure it out later today.

ldionne:
Ok, awesome. Just create a PR whenever you're ready.

brunocodutra:
Ok, will do.

brunocodutra:
Almost got it working for "Ninja", "NMake Makefiles", "Unix Makefiles" and "MSYS Makefiles", however:

* NMake hangs while compiling (progress bar never increases)
* Unix Makefiles produce the following error which I believe is related to Ruby

> Scanning dependencies of target path.to.benchmark
> [ 50%] Generating path/to/benchmark.json
> -e:2: unknown regexp options - tabch 
> CMakeFiles/path.to.benchmark.dir/build.make:64: recipe for target 'path/to/benchmark.json' failed

Any ideas?

brunocodutra:
...never mind Unix Makefiles, got it finally working by double quoting ruby commands.

brunocodutra:
...and out of the sudden NMake started working as well.

Alright, I'll squash it all to a single commit and open a PR later today.

ldionne:
Did you rebase your work on top of master? There was an error related to `-e:2: unknown regexp options - tabch ` when I worked on the initial support for Appveyor, but it's been fixed now.

brunocodutra:
I think so. I suppose that's when you double quoted one of the commands? I just did the same for all of them and everything started working.

PR #20, opened 2016-03-11
test other makefile generators on appveyor

brunocodutra:
following #18

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,17 +9,20 @@ branches:
   except:
     - /pr\/.+/
 
-os:
-  - Visual Studio 2015
-
 build:
   verbosity: detailed
 
-configuration:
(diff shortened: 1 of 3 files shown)


brunocodutra:
why do we filter branches starting with `/pr/` on appveyor?

ldionne:
> why do we filter branches starting with /pr/ on appveyor?

It's just for me. When I push a branch and create a pull request, two GitHub hooks are called; one because I pushed a branch, and one because I opened a PR. This is annoying, as we'd like the testing to be done only once. So I use `pr/...` and then the branch won't be tested, only the pull request.

Don't use `pr/...` for your own branches (it does not hurt though).

brunocodutra:
I was just curious. Makes sense.

Issue #15, opened 2016-03-09
[RFC] Splitting benchmarks into datasets


ldionne:
Here's an idea I've been thinking about; I also started implementing it, but it's very rough right now. The idea is to allow benchmarks to be described in `CMakeLists.txt` files. For example, you would have a directory structure

```
path/
    to/
        benchmark/
            CMakeLists.txt
            file1.cpp.erb
            ...
            fileN.cpp.erb
```

The `.cpp.erb` files would be just the same as the current `.cpp` files, i.e. they would contain the code to be benchmarked in the form of ERB templates. The `CMakeLists.txt` file would look somewhat like this:

```cmake
add_dataset(file1 "(0..100).step(5)") # <- the name of the file and the range to benchmark
add_dataset(fileN "....)
add_benchmark(name_of_the_benchmark file1 ... fileN)
```

The `add_benchmark` call would add a target that runs the benchmarks and gathers all the datasets (a bit like we do right now), but the dependency between a benchmark and the `.cpp` files it "contains" would be stated explicitly in CMake. Ok, but what are the advantages to doing this?

1. We could allow setting CMake properties to each dataset (such as include directories, flags, etc...) in a finer-grained manner, while still being portable if we run the benchmarks through CMake (as suggested in #14).
2. Users wouldn't have to worry about how the charts are specified. They would only have to call the CMake commands and we would generate the proper JSON files for them.
3. We could provide different kinds of benchmarks, for example compile-time and compiler RAM usage. This could be achieved by adding options to `add_benchmark` (like `add_benchmark(name [COMPILE_TIME|RAM] file1 ... fileN)`). For each kind of benchmark, we would have a simple template that sets reasonable defaults for chart visualization (for compile-time the yAxis is seconds, and we can set the titles of the axis ourselves).
4. We could exclude datasets from a benchmark based on CMake conditions naturally, without resorting to `cmake_bool`.

The only downside I see to this approach so far is that it's slightly less flexible, in the sense that being able to write arbitrary Ruby code in the `chart.json` file is pretty damn flexible. If we take over the control of `chart.json` and generate it ourselves from CMake, then we're taking away some flexibility, for better or worse.

Any thoughts?

brunocodutra:
I guess all the flexibility one needs may be expressed directly in the .cpp files.

[...]

PR #21, opened 2016-03-12
Splitting benchmarks into datasets

brunocodutra:
When merged, this will address #15


brunocodutra:
Enough is enough of fighting with [MSYS' path conversion rules](http://www.mingw.org/wiki/Posix_path_conversion), so I gave up on is at least for now.

Now NMake intrigues me, the error seems to be quite trivial but I can't figure out how to fix it. 

> LINK : fatal error LNK1104: cannot open file 'CMakeFiles/std_decay.dir/_metabench/decay/std_decay.cpp.obj'

Any ideas?

ldionne:
Just an hypothesis; perhaps the `.cpp` file is not written to disk (because of a missing flush, say), and that causes the linker to fail because no `.obj` is emitted. And the compiler error would be swallowed because of some reason.

But before we try to make it work on all the generators, I think it would be more important to split this into small logical commits that build upon each other, so I can sort-of review them. Otherwise, this PR is like a complete rewrite of the whole thing, and I'm not sure I want to merge this as-is.

I see a couple of changes that could be made independently of the core of this PR, I'll comment on the corresponding lines. If we can just reduce the diff a bit, it'll be easier to review.

README.md
```diff
@@ -108,19 +100,14 @@ values of `n`, and will gather benchmark data for each of these values.
 The `.cpp` file will be compiled (to benchmark it) as if it were located
 in the directory containing the `.cpp.erb` file, so that relative include
 paths can be used. Furthermore, it will be compiled as if the `.cpp` file
-was part of a CMake executable added in the same directory as the call to
+were part of a CMake executable added in the same directory as the call to
```
ldionne: This change can be made independently (e.g. fix typos in README).

README.md
```diff
@@ -108,19 +100,14 @@ values of `n`, and will gather benchmark data for each of these values.
...
-But how does Metabench figure which `n` we should benchmark a file for? These
-values are specified in the `chart.json.erb` file. This file, as outlined above,
-controls how the `.cpp.erb` files are rendered and how the data can be visualized
-as a chart.
-
 TODO: Finish this
 
 This is it for basic usage! Note that the `example/` directory contains a
-fully working minimal example of using Metabench to create a benchmark.
+fully working example of using Metabench to create benchmarks.
```
ldionne: As above, this can be made independently.

metabench.cmake
```diff
@@ -258,8 +231,8 @@ file(WRITE ${MEASURE_RB_PATH}
 # benchmarks. The template is completed by filling it with the contents
 # of the corresponding JSON file and the required JS libraries.
 ################################################################################
-set(CHART_TEMPLATE_HTML_PATH ${CMAKE_CURRENT_BINARY_DIR}/_metabench/chart_template.html.erb)
-file(WRITE ${CHART_TEMPLATE_HTML_PATH}
+set(CHART_HTML_ERB ${CMAKE_CURRENT_BINARY_DIR}/_metabench/chart.html.erb)
```
ldionne: This, along with the other changes to this file below, can be made independently. Though I'm not sure removing the word `_PATH` is wise, because it clearly indicates that we're talking about a path as opposed to the content of a file.

ldionne:
If you could split the changes I identified into separate commits at the beginning of your PR, I could merge them without merging the whole PR. Then, you could rebase your PR on top of master, which would contain those changes after I've merged them.

brunocodutra:
Right, what's the best approach for me to do that?
I'm not much of a git geek, so I'm not really sure how to go about that.

metabench.cmake
brunocodutra: Oh this one was just for symmetry, as the other global variables were named after the filename along with its extensions. I also removed the `_template` since now it has `.erb` already as an extension.

ldionne: Ok, I agree.

brunocodutra: We could also do it the other way around and append `_PATH` to the other variables instead.

ldionne: I would actually prefer that, but it's a nitpick really.

ldionne:
> Right, what's the best approach for me to do that?
> I'm not much of a git geek, so I'm not really sure how to go about that.

That's going to be a tough one to explain remotely. Basically you'll want to use `git rebase --interactive` to edit your commits and end up with the independent changes at the beginning of your chain of commits. Give me a moment, I'll do it (and push to a different branch) and I'll try to show you what I'm doing.

brunocodutra:
> Give me a moment, I'll do it (and push to a different branch) and I'll try to show you what I'm doing.

Sounds great, thanks!

ldionne:
```bash
# First step, remove the merge commit (that's ugly)

# Create a new branch 'cleanup' based on 'master'
git checkout -b cleanup master

# Try to apply the commit 'split benchmarks into datasets' on top of the
# current branch. You should get a conflict when doing this, because the 
# merge commit that we're trying to remove used to be _after_ the commit 
# we are cherry-picking, and now it's _before_. But this is what we want,
# because we like to have a linear history as much as possible.
#
# To resolve the conflict, you'll have to edit the metabench.cmake file and 
# keep only the changes that you made in the commit we're cherry-picking.
# In general what you keep depends on the situation, but here keeping your
# changes only is the right thing to do.
git cherry-pick 8938646

# Once you're done with the conflict, add metabench.cmake and finish the
# cherry-picking.
git add metabench.cmake
git cherry-pick --continue

# Ok, if you do `git log` now, you should see that we've effectively "rebased"
# the first commit of your pull request on top of 'master', and now the merge
# commit is not necessary. We'll now get the rest of the commits in your PR
# on top of this:
git rebase --onto cleanup dd8f2ec brunocodutra-feature/split

# Now, if you look at your history, you should see that brunocodutra-feature/split
# has the same content as your PR, except the merge commit has been removed.
# You don't need the 'cleanup' branch anymore, so remove it.
git branch -d cleanup


# Ok, now's time to actually get the changes we want into separate patches.
# For this, we'll use git rebase --interactive. This will drop you into your
# $EDITOR. You'll want to edit the commits that contain changes that you'd
# like to separate from the rest (use the 'edit' command in the interactive
# rebase). Here, I'll be editing 1bb5af7 (the README changes) only.
git rebase --interactive master

# Now, I have the independent changes to the README in a separate commit,
# and I can push this to a new branch. What I would then do is then merge
# this commmit to master independently, and let you rebase your pull request
# using `git rebase master` to include this new commit.
```

ldionne:
The resulting branch is https://github.com/ldionne/metabench/commits/brunocodutra-feature/split

ldionne:
Or the other approach; I'll merge #22 and then you'll rebase on top of master using `git rebase master feature/split`, fixing conflicts as you go. Probably simpler :-)

brunocodutra:
wow, that's thorough.
The minor changes to README and the renaming of that variable I have actually done manually meanwhile and submitted as a new pull request. 

I'll let you merge it, than I'll follow your tutorial to rebase this one on top of the new master.

ldionne:
I just merged #22. Also note that there ought to be a better way of removing the merge commit than what I did above (that's pretty complicated), but I wasn't able to do it so I went with cherry-picking.

brunocodutra:
Alright, it should be cleaner now.

metabench.cmake
```diff
@@ -6,127 +6,157 @@ cmake_minimum_required(VERSION 3.0)
 
 find_package(Ruby 2.1)
 if(NOT ${RUBY_FOUND})
-    message(WARNING "Ruby >= 2.1 was not found; the metabench.cmake module can't be used.")
-    return()
+  message(WARNING "Ruby >= 2.1 was not found; the metabench.cmake module can't be used.")
```
ldionne: The indentation here was changed to 2 tabs instead of 4. This is inconsistent with the rest of the file, so I suggest removing it.

metabench.cmake
```diff
@@ -6,127 +6,157 @@ cmake_minimum_required(VERSION 3.0)
...
 execute_process(COMMAND ${RUBY_EXECUTABLE} -r ruby-progressbar -r tilt -e ""
                 RESULT_VARIABLE __MISSING_GEMS
                 OUTPUT_QUIET ERROR_QUIET)
 if(${__MISSING_GEMS})
-    message(WARNING
-        "The 'ruby-progressbar' and/or 'tilt' gems were not found; "
-        "the metabench.cmake module can't be used."
-        "Use `gem install ruby-progressbar tilt` to install the missing gems.")
-    return()
+  message(WARNING
```
ldionne: Same comment as above regarding the indentation.

metabench.cmake
brunocodutra: Yes, my bad, I'm rebasing again right now to fix this among other things.

ldionne: No problem, that's what reviews are for.

brunocodutra:
I reverted any changes I'd made to `run_tests.sh` since they were actually orthogonal to splitting datasets.

brunocodutra:
I'll rebase atop master due to #24.

ldionne:
Ok. BTW, just `git rebase master` should work, no need to do the fancy dance with cherry-pick I showed you. You'll have a couple of conflicts, but nothing major (I've done it locally to continue reviewing your PR).

brunocodutra:
> Ok. BTW, just git rebase master should work

Cool, just a sec.

ldionne:
Here are some comments/issues with the PR as it currently stands:

- We do not output debug information when the benchmark fails. This can be fixed by reintroducing the `error_string` thing that was there and launching the compilation through `Open3.capture3`.

- On Ninja, the progress bar is only shown at the end, once the benchmark is completed. I believe this is because the terminal output is buffered by Ninja to avoid it being intermixed if we were to run many benchmarks at the same time. I'm not sure whether this should be addressed, but I'm pointing it out.

- `metabench_add_benchmark` now adds the target to the ALL builtin target. I believe this is undesirable, or at least it should be possible not to add it. A lot of times, the benchmarks are just optional. I suggest we remove this from the current PR, and address it in a separate issue (see #26).

ldionne:
Another thing:

- Right now, I don't think we can exclude a dataset from the benchmark based on a CMake condition. Or it might be possible, but not easy. I think this should be made possible (IDK how, though), because that was one of the reasons for splitting a benchmark in datasets in the first place.

brunocodutra:
> We do not output debug information when the benchmark fails. This can be fixed by reintroducing the error_string thing that was there and launching the compilation through Open3.capture3.

This is my first experience with ruby, so I really had no idea how to handle the possibility of compilation errors. Looking at your original implementation, would it suffice to just inset the following? (I'm not even sure what it means, I just copied over from your original implementation).

```
stdout, stderr, status = Open3.capture3(*command)

error_string = <<-EOS
compilation error: #{command_line.captures[0]}

stdout
#{'-'*80}
#{stdout}

stderr
#{'-'*80}
#{stderr}

code
#{'-'*80}
#{code}
EOS
raise error_string if not status.success?
```

> On Ninja, the progress bar is only shown at the end, once the benchmark is completed. 

Is there any way we could flush the progress bar?

> metabench_add_benchmark now adds the target to the ALL builtin target. I believe this is undesirable, or at least it should be possible not to add it. A lot of times, the benchmarks are just optional. I suggest we remove this from the current PR, and address it in a separate issue (see #26).

Targets may now be freely configured by the user, so it would just be a matter of issuing `set_target_properties`. At any rate it should be very easy to add an option ALL, so as to make it simpler.

> Right now, I don't think we can exclude a dataset from the benchmark based on a CMake condition. Or it might be possible, but not easy. I think this should be made possible (IDK how, though), because that was one of the reasons for splitting a benchmark in datasets in the first place.

Sure you can, you just don't add that dataset to the benchmark, like this:

```
set(datasets)

metabench_add_dataset(my_tuple msvc_tuple.cpp)
list(APPEND datasets my_tuple)

if(SOME_LIB_FOUND)
  metabench_add_dataset(some_lib_tuple some_lib_tuple.cpp)
  list(APPEND datasets some_lib_tuple)
endif
if(SOME_OTHER_LIB_FOUND)
  metabench_add_dataset(some_other_lib_tuple some_other_lib_tuple.cpp)
  list(APPEND datasets some_other_lib_tuple)
endif

metabench_add_benchmark(tuples DATASETS ${datasets})
```

ldionne:
> Looking at your original implementation, would it suffice to just inset the following? (I'm not even sure what it means, I just copied over from your original implementation). [...]

Almost. Actually, the following should work if you insert it in place of `time = Benchmark.realtime{...}`:

```ruby
time = Benchmark.realtime {
  stdout, stderr, status = Open3.capture3(*ARGV)
  error_string = <<-EOS
compilation error: #{ARGV.join(' ')}

stdout
#{'-'*80}
#{stdout}

stderr
#{'-'*80}
#{stderr}

code
#{'-'*80}
#{code}
EOS

  raise error_string if not status.success?
}
```


> Is there any way we could flush the progress bar?

I don't think we can force Ninja to output anything if it does not want to. Let's defer this to an issue when we'll merge this PR. I want to take small bites.


> Targets may now be freely configured by the user, so it would just be a matter of issuing set_target_properties. At any rate it should be very easy to add an option ALL, so as to make it simpler.

I see. I still think it would be better if this PR did not add the target to ALL, and then we can change it after. Again, this is for the sake of doing small changes as much as possible.



> Sure you can, you just don't add that dataset to the benchmark, like this: [...]

Ah! You're right, then this is not an issue.

brunocodutra:
Benchmarks are no longer added to ALL by default, but may be added explicitly to it using the following:

```{cmake}
set_target_properties(<target> PROPERTIES EXCLUDE_FROM_ALL FALSE)
```

ldionne:
Ok, good. We're not too far away. Another thing: when we set properties on a dataset target, the CTest target associated to it will not get that property. Therefore, I think it would be better to use the same target for both testing and the actual benchmark (i.e. no `target` and `target.test`, only `target`). We could use the value of `ENV['METABENCH_TEST_ONLY']` as we used to do, or something else if you have a better idea.

ldionne:
Note: I pushed some commits [here](https://github.com/ldionne/metabench/commits/ldionne-feature/split). I suggest you add them to your pull request.

brunocodutra:
> I think it would be better to use the same target for both testing and the actual benchmark (i.e. no target and target.test, only target)

That doesn't work very well, because the test target would effectively build the dataset, which won't be deemed outdated by CMake when we try to build a benchmark associated with that dataset and so one of two things happen :

* the benchmark will gather incomplete data from datasets which have been tested before;
* we don't produce benchmark data when testing a dataset (by checking `METABENCH_TEST_ONLY`) and we get an exception from ruby when it tries loading benchmark data.

ldionne:
You're right, using the same target causes problems. It might be possible to hack around this by touching or removing some file so that CMake considers the dataset as outdated, but then we'll always need to re-generate everything. I'll try to think of something tomorrow.

brunocodutra:
By the way, on a related topic, I couldn't get the dataset target to actually depend on its template (.erb) file, so that it gets rebuilt if the file is altered.

I tried including it as a source file and even creating a custom target that depends on it, but to no avail :-(

ldionne:
Since this is a rather large PR, I made a non-exhaustive checklist of functionality that we'd like to have or preserve from the current code:


- [x] Being able to pass an absolute or a relative path to `add_dataset`
- [x] Get the command line used to compile from CMake in order to be portable
- [x] Being able to include local headers
- [x] When a dataset or the `chart.json` file is modified, the HTML and JSON files containing the chart should be considered outdated.
- [ ] When the `.cpp.erb` is modified, the corresponding dataset should be considered outdated. I raised this issue above, but it would be fine to fix this in a subsequent PR.
- [ ] Targets for generating the HTML and JSON chart files (created with `add_benchmark`) should appear in the list of targets shown by `ninja help`. This is nice to have, but it could be addressed in a subsequent PR.
- [ ] The progressbar should be updated interactively using the Ninja generator. This is nice to have, but we should consider this in a subsequent PR too.
- [ ] Properties set on the dataset executable should also apply to the test executable. It would be nice to fix this before we merge, because otherwise we'll be merging something with a bug in it.
- [ ] We should be able to measure the size of the generated executable. As long as we _know_ we'll be able to address this later, then I'm fine with merging this PR without support for it yet. However, since we hijack the `RULE_LAUNCH_COMPILE`, I'm not sure how we might do that.

brunocodutra:
> Targets for generating the HTML and JSON chart files (created with add_benchmark) should appear in the list of targets shown by ninja help

Do you mean independently? Currently the benchmark target is responsible for generating both, would it suffice then to create two independent targets named `${target}.json` and `${target}.html` on which `${target}` would depend?

> We should be able to measure the size of the generated executable.

This could be accomplished by changing the `data:values:` attribute to be an array of triples instead of pairs, adding the size as a thid element that is. We could then get NVD3 to display [a line chart with two y axis](https://nvd3-community.github.io/nvd3/), left for time the right for size. I think however, this would be better addressed in a PR of its own, what do you think?

ldionne:
> > Targets for generating the HTML and JSON chart files (created with add_benchmark) should appear in the list of targets shown by ninja help

> Do you mean independently? Currently the benchmark target is responsible for generating both, would it suffice then to create two independent targets named ${target}.json and ${target}.html on which ${target} would depend?

Forget about this, it's fine as it is right now IMO. I'm just blind and I did not see the targets when I issued `ninja help`.



> > We should be able to measure the size of the generated executable.
>
> This could be accomplished by changing the data:values: attribute to be an array of triples instead of pairs, adding the size as a thid element that is. We could then get NVD3 to display a line chart with two y axis, left for time the right for size. I think however, this would be better addressed in a PR of its own, what do you think?

What I'm concerned about is how to gather the size of the executable. Since we're hijacking the `RULE_LAUNCH_COMPILE` property, we do not have access to the generated executable (since there's no executable). All there is is an object file, but no executable before the linker has finished is job. IIUC, what happens right now is

1. CMake calls us with a command line to compile the dataset file (and generate a `.o` from that). Since that's our hook, we actually run the compiler `n` times with `n` different `.cpp` files generated from the ERB template, and generate the JSON file containing benchmark data. The linker has not run once so far, and no executable was ever created, only object files.

2. CMake assumes that we have generated an object file, and calls the linker on the object file. We have indeed generated an object file (actually `n` object files but we overwrite them each time). The object file that CMake will call the linker on and generate an executable from is the last object file we generated.

Thus, I don't think there's a way with our current way of hooking into the compilation to measure the size of the generated executable.

brunocodutra:
> Thus, I don't think there's a way with our current way of hooking into the compilation to measure the size of the generated executable Edit: (for all different values of n).

Right, but wouldn't it be just the same to compare the sizes of the object files? I mean, for all that matters, the impact in footprint due to compiling for increasingly larger datasets should already  be observable through the size of the object file, or am I missing anything?

ldionne:
Link-time optimization. It can strip symbols and make a __huge__ object file reasonable again. That's not a concern for Metal, which is type-level only, but for Hana executable size are a real concern.

ldionne:
Also, people with size constraints will want the benchmarks to represent what they actually get (the executable), not the object file, and I can understand that.

brunocodutra:
> Link-time optimization. It can strip symbols and make a huge object file reasonable again. That's not a concern for Metal, which is type-level only, but for Hana executable size are a real concern.

Makes sense.

> Also, people with size constraints will want the benchmarks to represent what they actually get (the executable), not the object file, and I can understand that.

I thought about that too actually.

I guess then we would have to also hijack the [`RULE_LAUNCH_LINK`](https://cmake.org/cmake/help/v3.0/prop_gbl/RULE_LAUNCH_LINK.html) property to run a loop linking each object file so we may get the size of the executable that is generated. We could achieve this by renaming every object file right after compilation so that they don't get overwritten and then, at link time, we rename them back, one at a time, prior to linking.

ldionne:
Wouldn't it be simpler to keep it the other way around, and have our script run CMake and parse the output to get the timings that our `RULE_LAUNCH_COMPILE` outputs (what we do right now on master)? It seems like we would be fighting less against CMake. It would also solve our problem of depending on the `.cpp.erb` template, because then the custom command that generates each `.json` dataset can just depend on it. The only thing that this does not solve is the sleeping issue, but it might be possible to solve it some other way. But don't change anything for now; I'll try to see how hard it would be to incorporate your design of split datasets with the current way of running the benchmarks.

ldionne:
Ah crap, if we keep the current way then we can't add properties to the target that we use to represent the dataset, because that target will be some `custom_target` instead of an `add_executable`. I'm still looking, but there's a dilemma there.

brunocodutra:
> Wouldn't it be simpler to keep it the other way around, and have our script run CMake and parse the output to get the timings that our RULE_LAUNCH_COMPILE outputs (what we do right now on master)?

To be honest I find it less hacky (read creative) to override the `RULE_LAUNCH_LINK` and I see some drawbacks:

* If several benchmarks reuse the same dataset (which should be just fine once we split datasets), it would have to be compiled all the way for all `n` once per benchmark.
* I don't see a clean way to make `metabench_add_benchmark` aware of the `.erb` template used to create each dataset.
* The dataset itself wouldn't use the `.erb` template at all so in this sense it would be rather artificial to associate a dataset with a specific file.

brunocodutra:
> Ah crap, if we keep the current way then we can't add properties to the target that we use to represent the dataset, because that target will be some custom_target instead of an add_executable. I'm still looking, but there's a dilemma there.

Actually I though a dataset would be a generic executable target which would compile a source file provided on the fly by the benchmark target.

ldionne:
I'm not sure we understand each other. To be sure, what I'm thinking about is something like the following (pseudo-CMake):

```cmake
# Note: After using add_dataset, users can refer to `target` as an executable 
#       to which properties can be added.
function(add_dataset target path/to/template.cpp.erb range)
    # Add a dummy executable that we will use to collect the timings.
    file(WRITE path/to/template.cpp "")
    add_executable(${target} path/to/template.cpp 
        RULE_LAUNCH_COMPILE time compilation and print timing to stdout
        RULE_LAUNCH_LINK time linking (optional), check exe size and print to stdout
    )

    # Add a command to generate the JSON file that will contain the timing data.
    add_custom_command(OUTPUT path/to/template.json
        COMMAND for each n  in 'range', 
                - generate the .cpp file from the ERB template
                - write it to path/to/template.cpp 
                - run the ${target} above and parse stdout to get statistics
                - write those statistics to path/to/template.json
    )

    # Add the test target 
    add_test(NAME ${target} 
        COMMAND for some n in 'range',
                - generate the .cpp file from the ERB template
                - write it to path/to/template.cpp
                - run the ${target} above
    )
endfunction()

function(add_benchmark target datasets)
    # Basically the same as you
endfunction()
```

The only real change is that instead of doing the `n` different compilations within the same run of `${target}`, we would do `n` different calls to `${target}`, each of which would time exactly one compilation.

brunocodutra:
> I'm not sure we understand each other. To be sure, what I'm thinking about is something like the following [...]:

Oh now I think get it, I guess this would work fine.

brunocodutra:
Meanwhile I believe I found [a way](https://cmake.org/cmake/help/v3.0/prop_tgt/INTERFACE_COMPILE_DEFINITIONS.html#prop_tgt:INTERFACE_COMPILE_DEFINITIONS) to solve the problem of having targets share compile options:

> Consuming targets can add entries to their own `COMPILE_DEFINITIONS` property such as `$<TARGET_PROPERTY:foo,INTERFACE_COMPILE_DEFINITIONS>` to use the compile definitions specified in the interface of foo

So basically any `PUBLIC` property of `foo` can be inherited by `bar` using the above syntax.

ldionne:
Interesting. So that would fix the issue about synchronizing the properties of the test and the dataset targets. On my side, I've managed to implement something that seems to work using the approach I posted above. I'll create another PR so we can discuss.

brunocodutra:
> I'll create another PR so we can discuss.

Wouldn't it be better to rebase it atop this one?

ldionne:
Ok, here's what I'm going to do. I'll merge this PR, and then I'll create a new one which reintroduces the "old" way of running the benchmarks. Then we can discuss on that one. Works for you?

ldionne:
It's dirty but I'm tired of fighting GitHub's inappropriate tools for collaboration, so whatever.

brunocodutra:
Either way really, I just thought it would be nice to have the discussion all at one place, but since github lacks proper tools for collaborative PR there's not much we can do. 

Anyways, if you developed from master it should be quite distinct from this one here and would be a pain to rebase, so it would probably be easier to just open a new PR mentioning this one.

brunocodutra:
... comments should really update live.

ldionne:
lol, at least we're not paying for the service.

PR #24, opened 2016-03-13
Make sure we generate charts

### LINKED ISSUE #23: We currently do not output any HTML or JSON chart
The double-quoting added in 356855af57fa1640df4c8f4a020e8accc13dd972 had the effect of turning all the commands we were executing into Ruby strings. So we were evaluating the strings containing our code instead of the code itself.

ldionne:
This PR is an attempt to fix #23. I'm submitting as a PR instead of merging into master, because I think that might break some generators so I might still need to adjust some things.


brunocodutra:
As part of #21 I made some changes to `run_tests.sh` in order to check whether `check.json` and `check.html` are generated.

I'm cherry picking this into a new PR.

ldionne:
> I'm cherry picking this into a new PR.

No need, i'll merge these change. I'd rather not tie the `run_test.sh` too much to the actual tests, which is why I added the check through CMake.

brunocodutra:
> [...] which is why I added the check through CMake.

Much better.

ldionne:
@brunocodutra I realize that merging this might give you some conflicts when you'll rebase #21. However, recovering basic functionality is, I think, the number 1 priority for now.

brunocodutra:
> I realize that merging this might give you some conflicts when you'll rebase #21. However, recovering basic functionality is, I think, the number 1 priority for now.

Of course, no problem.

PR #27, opened 2016-03-15
Generate datasets by running behind CMake

ldionne:
Note that this fixes all the issues I mentioned in [this comment](https://github.com/ldionne/metabench/pull/21#issuecomment-196445287) of #21.


metabench.cmake
```diff
@@ -44,54 +50,68 @@ set(METABENCH_DIR ${CMAKE_CURRENT_BINARY_DIR}/_metabench)
...
+#       must be provided as a Ruby Iterable (such as `(1...100).step(5)`).
+#
+# [1]: http://nvd3.org/
+function(metabench_add_dataset target path_to_template range)
+    # Transform any absolute path to a relative path from the current source directory.
+    if (IS_ABSOLUTE ${path_to_template})
+        file(RELATIVE_PATH path_to_template ${CMAKE_CURRENT_SOURCE_DIR} ${path_to_template})
+    endif()
+    get_filename_component(path_to ${path_to_template} DIRECTORY)
+    get_filename_component(template ${path_to_template} NAME_WE)
```
brunocodutra: Much better!

metabench.cmake
```diff
@@ -44,54 +50,68 @@ set(METABENCH_DIR ${CMAKE_CURRENT_BINARY_DIR}/_metabench)
...
-        COMMAND ${CMAKE_COMMAND} --build . --target ${target}.test
-        WORKING_DIRECTORY "${PROJECT_BINARY_DIR}"
+
+    # We also setup a CTest target to test the generation of the dataset.
+    # We do not actually collect any data here.
+    add_test(NAME ${target}
+        COMMAND ${CMAKE_COMMAND} -E env METABENCH_TEST_ONLY=true
+            ${RUBY_EXECUTABLE} -r tilt/erb -r ${configured_metabench_rb}
+            -e "range = (${range})"
+            -e "data = Metabench.compile_time(\"${CMAKE_CURRENT_SOURCE_DIR}/${path_to_template}\", range)"
```
brunocodutra: wouldn't it be much easier to just pass `[range[0], range[-1]]` to Metabench.compile_time?

metabench.cmake
```diff
@@ -155,7 +175,7 @@ function(metabench_add_benchmark target)
                 .render(nil, {chart: chart.to_json, nvd3_css: nvd3_css, nvd3_js: nvd3_js, d3_js: d3_js})"
             -e "IO.write('${json_path}', chart[:data].to_json)"
             -e "IO.write('${html_path}', index)"
-        DEPENDS ${ARGS_DATASETS} "${ARGS_CHART}"
+        DEPENDS ${data} "${ARGS_CHART}"
```
brunocodutra: good catch

metabench.cmake
```diff
@@ -172,78 +192,115 @@ endfunction()
...
+"require 'ruby-progressbar'                                                                             \n"
+"require 'tempfile'                                                                                     \n"
+"require 'tilt/erb'                                                                                     \n"
+"                                                                                                       \n"
+"                                                                                                       \n"
+"module Metabench                                                                                       \n"
+"  # This function is meant to be called inside a ERB-based `chart.json` file.                          \n"
+"  # `erb_template` should be the name of the `.cpp.erb` file containing the                            \n"
+"  # benchmark to run.                                                                                  \n"
+"  def self.measure(erb_template, range, env: {})                                                       \n"
```
brunocodutra: I don't think we need `env` anymore

metabench.cmake
```diff
@@ -172,78 +192,115 @@ endfunction()
...
+"    measure_file = Pathname.new('${METABENCH_DIR}/\@path_to\@/\@target\@.cpp')                         \n"
+"    exe_file = Pathname.new('${METABENCH_DIR}/\@path_to\@/\@target\@${CMAKE_EXECUTABLE_SUFFIX}')       \n"
+"    command = ['${CMAKE_COMMAND}', '--build', '${CMAKE_BINARY_DIR}', '--target', '\@target\@']         \n"
+"    range = range.to_a                                                                                 \n"
+"    range = [range[0], range[-1]] if ENV['METABENCH_TEST_ONLY'] && range.length >= 2                   \n"
```
brunocodutra: this would be much cleaner passed as argument (see above)

Ok I admit it, I'm allergic to METABENCH_TEST_ONLY =P

brunocodutra:
Since Metabench is about producing all kinds of benchmark data from some `.erb` template, I think it would be much better if the JSON file we output contained both aspects,  that is time and executable (currently). My suggestion is to output benchmark data as an array of triples of the form `[ ..., [n, time_n, size_n], [m, time_m, size_m], ...]`. 

Recapping our previous discussion in #21, more specifically [this comment](https://github.com/ldionne/metabench/pull/21#issuecomment-196456789), it shouldn't be hard to plot a chart containing two y axis if data were output in this format. 

I should point out that NVD3 should work as is right now, even if we start outputting triples, as it is set up to specifically look for the first and second elements of each element in the `values:` array in order to plot time benchmarks. Therefore we could safely change the format to triples now and then later we look into plotting multi axis charts in a separate PR (I'm on it).

metabench.cmake
ldionne: Thanks, fixed.

metabench.cmake
ldionne: Indeed, I hate that hack too and that was in my list of things to do. We shouldn't have to configure the `metabench.rb` file. Even more, we could make `metabench.rb` a Ruby Gem and list the only dependency of this project as being the  `metabench` gem.

brunocodutra: Sounds fancy.

No idea what that would mean though :I

ldionne: Basically, you'd have to `gem install metabench` and that would install both `metabench.rb` and any dependencies.

ldionne: It's easy to create a gem with Ruby, I'll look into it later maybe.

ldionne:
Yes, I think it makes sense that we should output as much data as we have in the JSON file, and then let the HTML chart pick whatever it wants. But that should be handled separately. Are you OK if I merge this once the tests pass?

metabench.cmake
brunocodutra: Cool, but then how does `gem install` know where to fetch `metabench` from?

ldionne:
Your comment seems to have disappeared.

> Cool, but then how does gem install know where to fetch metabench from?

We would register metabench on rubygems.org. But actually, none of this will be needed because I just got rid of all the dependencies locally (except Ruby of course). I'll push the changes once this is merged.

brunocodutra:
> Your comment seems to have disappeared.

Which one you mean?

> But actually, none of this will be needed because I just got rid of all the dependencies locally (except Ruby of course)

That's awesome!

>  Are you OK if I merge this once the tests pass?

Sure.

ldionne:
> > Your comment seems to have disappeared.
>
> Which one you mean?

IDK, I received an email from GitHub saying that you had commented 

> Cool, but then how does gem install know where to fetch metabench from?

on a diff that no longer existed. No big deal.

metabench.cmake
```diff
@@ -170,80 +194,112 @@ endfunction()
...
-"  }                                                                         \n"
-"  datum[:values] << [n, time]                                               \n"
-"  progress.increment                                                        \n"
-"  source_file.write('')                                                     \n"
-"end                                                                         \n"
-"IO.write('\@datum\@', datum.to_json)                                        \n"
+"      raise error_string if not status.success?                                                        \n"
+"      data[:time] = stdout.match(/\\[compilation time: (.+)\\]/i).captures[0].to_f                     \n"
+"      # Size of the generated executable in KB                                                         \n"
+"      data[:size] = File.size(exe_file).to_f / 1000                                                    \n"
```
brunocodutra: shouldn't we express the executable size in base two rather than base ten, so this would be 1024?

ldionne: Lol, neither do I. Let's leave it in base 10 unless there's a reason to change it.

PR #28, opened 2016-03-15
Remove all external Ruby dependencies

ldionne:
Most of the changes are trivial, but I'd like to have thoughts on the format of the progress output. Right now, it's going to output something like

```
5/5 /Users/ldionne/code/metabench/example/decay/std_decay.cpp.erb (n = 400)
```

I think that's OK, but if you have a better idea then I'm open for suggestions (as long as it's easy to implement!).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -24,16 +24,10 @@ install:
   - if "%CMAKE_GENERATOR%" equ "Unix Makefiles" (set "PATH=C:\cygwin\bin;%PATH%")
   - if "%CMAKE_GENERATOR%" equ "MSYS Makefiles" (set "PATH=C:\MinGW\bin;C:\MinGW\msys\1.0\bin;%PATH%")
 
-  ############################################################################
-  # Use a more recent Ruby by default, and install required gems
-  #
-  # Note that we set the RUBY_LIBRARY variable explicitly because CMake fails
-  # to find it otherwise.
-  ############################################################################
+  # Use a more recent Ruby by default. We need to set the RUBY_LIBRARY variable
+  # explicitly because CMake fails to find it otherwise.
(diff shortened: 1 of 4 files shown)


brunocodutra:
I usually preffer displaying relative paths to reduce clutter.
Also, it would be nice if we displayed benchmark data after each step.

ldionne:
> I usually preffer displaying relative paths to reduce clutter.

Good point, fixed.



> Also, it would be nice if we displayed benchmark data after each step.

I'm not sure I understand. Do you mean that at each step we should output the compilation time and executable size? If so, then we can't do it for the current iteration, because the data is not available yet when we update the progress. We could show the data of the previous iteration, but IDK how useful that would be.

brunocodutra:
> I'm not sure I understand

I don't have CMake, nor ruby, nor, in fact, anything at hands right now but a stupid browser, so I can't see how it works, but I reckon each iteration would be displayed on a new line, like this:

```
1/5 [...]
2/5 [...]
...
5/5 [...]
```

So I thought it would be cool to display compilation time and executable size right after each iteration, like this:

```
1/5 [...]
elapsed time: x.xx s
executable size: yyy kb
2/5 [...]
elapsed time: x.xx s
executable size: yyy kb
...
5/5 [...]
```

Just so that the user may reduce turnaround time when trying to improve performance.

ldionne:
Oh, sorry, no actually the progress is updated live like for the Ninja generator. There's always only a single line displaying the progress.

ldionne:
If you look at the [Travis build](https://travis-ci.org/ldionne/metabench/jobs/116220435#L204), you'll see that there's only one line reporting the progress.

brunocodutra:
Got it now, forget what I said, it looks great as it is.

ldionne:
Awesome. Merging.

Issue #25, opened 2016-03-13
Add testing of different Makefile generators on Appveyor


ldionne:
The testing on many generators that was added in #20 was reverted in #24 in order to fix bug #23. So, right now, only the Ninja generator is tested on Appveyor.

ldionne:
Closed by #46. We still don't support NMake and Unix Makefiles on Windows, but we'll only invest time in this is someone shows interest for it.

PR #29, opened 2016-03-15
Re-enable exotic generators on Appveyor

ldionne:
This PR is an attempt to address #25 once and for all.


brunocodutra:
In light of #38, we should replace "MSYS Makefiles" by "MinGW Makefiles".

Notice however it complaints about `sh.exe` being in `PATH`, so be sure to add the following:

```{bash}
- set PATH=C:\MinGW\bin;%PATH:C:\Program Files\Git\usr\bin;=%
```

brunocodutra:
Any idea what causes NMake to hang?

ldionne:
> Notice however it complaints about sh.exe being in PATH, so be sure to add the following:

I don't see this complaint?



> Any idea what causes NMake to hang?

No, I thought I would figure it out after the rest works, since this one seems obscure.

brunocodutra:
> >  Notice however it complaints about sh.exe being in PATH, so be sure to add the following:
> 
> I don't see this complaint?

Not sure if we understand each other, I warned about the fact that using "MinGW Makefiles" on appveyor doesn't work out of the box because `sh.exe` is in `PATH`. It is put there by the git installation, so we gotta manually remove it prior to calling CMake.

Also "Unix Makefiles" and "MSYS Makefiles" must be used within cygwin/msys, which we don't do anymore, so I perhaps we should remove these.

ldionne:
Sorry, but I really don't understand what you mean. Shouldn't Unix Makefiles and MSYS Makefiles work provided we set the path properly as you initially set up? I'll try to figure why NMake is hanging instead.

brunocodutra:
> Sorry, but I really don't understand what you mean. Shouldn't Unix Makefiles and MSYS Makefiles work provided we set the path properly as you initially set up?

These can only be used within a posix environment, so, on Windows, that means from within cygwin and msys respectively. Previously we used `bash` to drive unit tests, so that's why they [kindda] worked, but now that we drive our tests directly from within `cmd` CMake is unable to use them.

We have thus two options:

1.  Set a variable `SHELL` to either `cmd`, `cygwin/bash` or `msys/bash` accordingly and prefix every call to CMake by `- call %SHELL%`.
2. Forget about Unix/MSYS Makefiles on Windows, because, well, we are on Windows.

That aside, now that we don't use a shell script to drive our tests, we may test MinGW Makefiles, which may only be used within Windows native shell. It will error out if used within a posix environment and that should explain why it complains about `sh.exe` being in `PATH`.

ldionne:
Ok, I found out why the NMake were going into an infinite loop. See the explanation in 7324551.

Now, regarding this whole "support fancy generators" thing; I actually only need and care about Unix Makefiles and Ninja on non-Windows platforms. As you can probably see, I have very little experience and knowledge about Windows and anything related, and I don't use the platform. Thus, what I suggest is that we drop support for anything too complicated to set up easily on Appveyor.

I have nothing against supporting more than the strict minimum on Windows, but I am the worst person to implement such support, so I'll drop this PR. If anyone needs to use other generators on Windows, then I'll gladly accept PRs and I'll provide the help I can, but for now I feel like we might be implementing something that nobody will ever use.

brunocodutra:
> Thus, what I suggest is that we drop support for anything too complicated to set up easily on Appveyor.

I agree, that's why I suggested dropping support for Unix Makefiles and MSYS Makefiles on Windows. Also, the only generator that seems native on that platform is NMake, so, In that sense, I'd say that supporting NMake would be the strict minimum on Windows.

> If anyone needs to use other generators on Windows, then I'll gladly accept PRs and I'll provide the help I can, but for now I feel like we might be implementing something that nobody will ever use.

Right, I'll look into that.

PR #33, opened 2016-03-16
Unify benchmark data

brunocodutra:
Currently we only export one benchmarking aspect, even though we gather data regarding multiple aspects (currently compilation time and executable size). This PR shall address this issue by making sure all available data is exported to the user.

**Note:** this is based on PR #32


brunocodutra:
Currently we only provide visualization for compilation time, but it should be possible to control which aspects are displayed.

One idea is to create a new property `aspect` in the `chart.json`, which defaults to `[time]` that may be used to select the aspects to be plotted. This would required a NVD3 `multiChart` instead of the simpler `lineChart`, but I think that's hardly an issue.

What do you think?

ldionne:
Providing the ability to select which aspect is to be plotted is a must. However, I'm not sure about the value of displaying more than one of them at the same time on a single chart (if I understand your proposal correctly). In all cases, I think this PR is good since it only makes more stuff available in the JSON file, which is good regardless of what we do with it.

brunocodutra:
> Providing the ability to select which aspect is to be plotted is a must. However, I'm not sure about the value of displaying more than one of them at the same time on a single chart

Hmm I myself don't have a use for it either right now, so I think I'll take the shortest path and just provide a way for select which single aspect to display.

ldionne:
Works for me. Let's not over engineer.

Issue #35, opened 2016-03-16
Consider adding a suite of metaprogramming benchmarks


ldionne:
The primary reason why we're writing this tool, I think, is because we all need some sort of benchmarking for our own metaprogramming library. Furthermore, we'll necessarily end up benchmarking our library against other people's libraries. This leads to a duplication of effort if we all do the same in our own sandbox.

Instead, what we could do is write a suite of benchmarks for many different libraries, once, and store these benchmarks in this repository (or another repository). These benchmarks would thus be centralized in a single place, and the duplication of effort would be avoided. A non-exhaustive list of libraries that might want to participate:

- [MPL](https://github.com/boostorg/mpl)
- [Fusion](https://github.com/boostorg/fusion)
- [Brigand](https://github.com/edouarda/brigand)
- [Hana](https://github.com/boostorg/hana)
- [Metal](https://github.com/brunocodutra/metal)
- [Meta](https://github.com/ericniebler/meta)

This is just an idea, I'm not sure whether it's a good one.

@brunocodutra @edouarda @ericniebler Is this something that would be useful to you?

edouarda:
I think it's a great idea!

ericniebler:
Peter Dimov has a couple of blog posts where he benchmarks some meta programs. We could build off of his work.

ldionne:
Actually, Hana already has a ton of these benchmarks written in a way compatible with Metabench, so I would suggest we start from there if anything. See #36 for an example of how I see this. Basically to start off, what I would need would be for everyone to provide a FindXXX CMake module that can be used to fetch their library (I already have FindHana, FindBoost and FindMeta, but I need FindMetal and FindBrigand if those exist).

brunocodutra:
Excellent way for us to bring our efforts together towards better
metaprogramming.

Metal already provides MetalConfig.cmake and MetalConfigVersion.cmake upon
installation, which enable the use of `find_package`. Basically you just
need to install Metal somewhere and tell CMake about it by adding the
installation path prefix to `CMAKE_PREFIX_PATH`, in case it is not a
default location that is.

ldionne:
Ok, it seems like we'll go forward with this idea so I'll close this issue in favour of PR #36.

PR #36, opened 2016-03-16
Initial draft of a benchmark suite

ldionne:
This is a rough example of how I would imagine a shared benchmark suite as discussed in #35. Hana has a _lot_ of these benchmarks already written, some for external libraries, so they could just be copied from there.


brunocodutra:
For some reason Meta doesn't get downloaded for me even though it is said to be found.

ldionne:
`cmake --build . --target Meta` should do the trick. But I think your approach with the MetalConfig file is more idiomatic, I should switch to that for Hana too.

brunocodutra:
> `cmake --build . --target Meta` should do the trick.

Oh I see! Interesting, so I guess if we add the target `Meta` as a dependency of its datasets, then Meta would get downloaded as necessary?

> But I think your approach with the MetalConfig file is more idiomatic 

And it seems it's a lot less typing too. However, now that I know how it works, I like Meta's approach for Metabench in particular. As long as we add the external projects as dependencies of their datasets, no intervention is required and we are also guaranteed to always benchmark the latest versions. That would be ideal for automating on Travis for example.

I've pushed a [commit](https://github.com/brunocodutra/metabench/commit/2ef0873e188f7ebed5dde132aaa26f2f29122428) to `pr/benchmarks` to make sure it is able to benchmark Metal, I suggest you add it to this PR.

I'd vote for merging this PR so that we may develop in parallel easier. Alternatively, we could coordinate for a little while, pulling each other's commits, till we can merge this to master, but I don't think there's a need for the extra effort.

ldionne:
I added your commit. But I think CMake's behaviour is really strange; if the MetalConfig.cmake file can't be found (because Metal was not installed), I get a big ugly warning from CMake. But it seems like this should work smoothly and simply report that Metal was not found, since this is exactly the purpose of these XXXConfig.cmake files?

ldionne:
Also, I would prefer to wait a bit before we merge this. I have a large amount of Hana benchmarks, and I would like to transfer them here before we start developing in parallel. The reason is that we'll want to benchmark the same thing for all metaprogramming libraries. Thus, we'll need a "canonical example" of what we are trying to measure for each benchmark. By taking the benchmarks in Hana, I can set that example.

brunocodutra:
By providing `XXXConfig.cmake` and `XXXConfigVersion.cmake` one enables a default `FindXXX.cmake` implementation provided by CMake as opposed to providing a custom implementation as you did for Hana.

It seems [`find_package`](https://cmake.org/cmake/help/v3.0/command/find_package.html) may be shut up using the option `QUIET`.

brunocodutra:
> Also, I would prefer to wait a bit before we merge this

Alright, we coordinate for a little while then. Later this evening I'll experiment with providing `FindXXX.cmake` for all libraries that we currently support based on our current implementation of `FindMeta.cmake`, so that it works out of the box with as little tweaking as possible.

ldionne:
Ah, thanks!

ldionne:
> Later this evening I'll experiment with providing FindXXX.cmake for all libraries that we currently support based on our current implementation of FindMeta.cmake, so that it works out of the box with as little tweaking as possible.

Hmmm, actually I think the better way to go would be for both Hana and Meta to provide XXXConfig.cmake files. Then, in the CI where we'd run the benchmarks, we would simply download the latest version of each library and install it (the installation should place the XXXConfig.cmake in the right location.

brunocodutra:
> I think the better way to go would be for both Hana and Meta to provide XXXConfig.cmake files

I agree in general, hence the way I set up Metal, but for metabench in particular I think it would be so much easier for the user to just download it and run make, no further concerns about the dependencies. It doesn't have to be set up via `FindXXX.cmake`, simply adding the external projects in the CMakesLists.txt in case `find_package` is unable to find a given library should do the trick. This way the user also has the option to override it so as to test against specific versions of a given library for example.

ldionne:
Ok, agreed.

ldionne:
Managing our dependencies is a challenge. Think of it; if we point to the most recent development version of each library, our benchmarks may break suddenly for no apparent reason, because of a bug/breaking change in one library. Perhaps that's desirable, perhaps not, but it has to be thought out.

Also, when would we run the benchmarks? We can do it from a Travis job, but when do we trigger that Travis job? That's a bit tricky..

Here's an idea: we could specify the version of each library on which we depend. By default, we'd fetch that version of the library, but it would be overridable somewhere in `benchmark/CMakeLists.txt` to make prototyping easier. Then, when a participating library releases a new version, they can submit a PR changing the version to which we're pointing. That PR would trigger a Travis job (as currently), and the benchmarks would be tested against the new library version in that job.

Finally; should we actually run the full benchmarks (instead of just the tests) and make the thus-gathered JSON data available in some way?

brunocodutra:
> Managing our dependencies is a challenge. Think of it; if we point to the most recent development version of each library, our benchmarks may break suddenly for no apparent reason, because of a bug/breaking change in one library. Perhaps that's desirable, perhaps not, but it has to be thought out.

It should definitely default to master or even to the latest release (I think there's an URL for that)

> Also, when would we run the benchmarks? We can do it from a Travis job, but when do we trigger that Travis job? 

[nightli.es](http://nightli.es/) - it's self explanatory

> Finally; should we actually run the full benchmarks (instead of just the tests) and make the thus-gathered JSON data available in some way?

Actually I was thinking about using `gh_pages` and nightli.es to display up to date benchmarks of all supported libraries.

ericniebler:
Thanks for taking this on, guys. Wishing I had more time to help.

ldionne:
No worries.

ldionne:
Ok, with the last commit, I fetch the latest version of each library on GitHub instead of using FindModules. Suggested roadmap:

1. Figure out how we're going to make the charts and their JSON data available to users (on `gh-pages`). Could be some kind of simple website, or something more barebones, IDK.
2. Setup Travis jobs that run the benchmarks nightly using nightli.es.
3. Finish exporting the benchmarks from Hana, and implement those benchmarks with other libraries.

brunocodutra:
Sounds good.

brunocodutra:
I've been pushing a few commits to `pr/benchmarks`.
Consider adding those to this PR.

ldionne:
I'm going to merge this even though it is still a work in progress so that we can both work on it more easily. However, please refrain from adding new benchmarks for now, for two reasons:

1. They might already be in https://github.com/boostorg/hana/tree/master/benchmark.
2. It's easier to tweak the framework while there are only a few benchmarks. If we add too many now, we'll lose a lot of time in editing stuff by hand every time we make changes.

brunocodutra:
> However, please refrain from adding new benchmarks for now, for two reasons:

Alright.

ldionne:
Can you please submit PRs for your commits:

1. One for adding `metal::transform`
2. One for the CMake-related changes inside `benchmark/`

brunocodutra:
Sure, I'll submit a PR for adding `type.reverse` as well, as it appears there's only the heterogeneous version in Hana.

(I did check that prior to adding)

ldionne:
Ok, my bad then, I thought the types were already there. Please do submit a PR for this one too, then.

brunocodutra:
No worries.

PR #38, opened 2016-03-18
Use a CMake file to drive the whole project

ldionne:
Instead of using a ad-hoc `run_tests.sh` file, this PR runs the unit tests through CMake. The reason for doing this is that it will ease the integration with benchmarks.

1. When we run the `all` target, the examples (and perhaps the benchmark suite when we're there) are built
2. When we run the "unit tests" with CTest, all the unit tests of Metabench are run, including the benchmark tests in the future


test/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+    add_custom_target(test.${dir}
+        COMMAND ${CMAKE_COMMAND} -E remove_directory ${dir}/build
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${dir}/build
+        COMMAND ${CMAKE_COMMAND} -E chdir ${dir}/build ${CMAKE_COMMAND} .. -G${CMAKE_GENERATOR} ${METABENCH_UNIT_TEST_OPTIONS}
+        COMMAND ${CMAKE_COMMAND} --build ${dir}/build --target check
+        COMMAND ${CMAKE_COMMAND} -E chdir ${dir}/build ${CMAKE_CTEST_COMMAND} --output-on-failure
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+        VERBATIM USES_TERMINAL
+    )
```
brunocodutra: Why setup a CMake target to call CMake to build another target?

ldionne: As explained in the comment above, I want each test to be isolated in its own sandbox. For example, not all the unit tests use `enable_testing`, so it is easier to define a unit test as being a self-contained CMake project than using `add_subdirectory` and having some subtle interaction between unit tests. Is this what you meant?

ldionne: Or did you mean why setup a `custom_target` and then call it from `add_test` as opposed to simply doing the work in `add_test`? If so, this is because `add_test` does not allow for multiple commands.

test/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+        continue()
+    endif()
+
+    add_custom_target(test.${dir}
+        COMMAND ${CMAKE_COMMAND} -E remove_directory ${dir}/build
+        COMMAND ${CMAKE_COMMAND} -E make_directory ${dir}/build
+        COMMAND ${CMAKE_COMMAND} -E chdir ${dir}/build ${CMAKE_COMMAND} .. -G${CMAKE_GENERATOR} ${METABENCH_UNIT_TEST_OPTIONS}
+        COMMAND ${CMAKE_COMMAND} --build ${dir}/build --target check
+        COMMAND ${CMAKE_COMMAND} -E chdir ${dir}/build ${CMAKE_CTEST_COMMAND} --output-on-failure
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
```
brunocodutra: It would be nice if we could do all the dirty work in the `CMAKE_CURRENT_BINARY_DIRECTORY` instead of polluting the source tree.

.appveyor.yml
```diff
@@ -35,4 +35,8 @@ before_build:
   - cd C:\projects\metabench
 
 build_script:
-  - bash tools\run_tests.sh -DRUBY_LIBRARY=%RUBY_LIBRARY%
+  - mkdir build
+  - cd build
+  - cmake .. -G%CMAKE_GENERATOR% -DRUBY_LIBRARY=%RUBY_LIBRARY% -DMETABENCH_UNIT_TEST_OPTIONS="-DRUBY_LIBRARY=%RUBY_LIBRARY%"
```
brunocodutra: What's up with this `RUBY_LIBRARY` thing, shouldn't we just make sure it's in `PATH`?

ldionne: It's super annoying. For some reason, the library can't be found (whether in path or not). I think it's a bug in the FindRuby module (like a name mismatch). At some point I just figured I could set `RUBY_LIBRARY` to shut it up, but I agree that we should fix the root of the problem. I'm looking into it right now.

test/CMakeLists.txt
ldionne: Good point, fixed.

.appveyor.yml
ldionne: Let's move this to #39.

ldionne:
Ok, I'm merging this. The remaining problem with `RUBY_LIBRARY` will be fixed in #39.

PR #39, opened 2016-03-18
[Appveyor] Properly find Ruby

ldionne:
Right now, the FindRuby module is unable to find the ruby library. To workaround this, we currently set the `RUBY_LIBRARY` CMake variable explicitly in `.appveyor.yml`. This PR is an attempt to fix this.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -24,11 +24,9 @@ install:
   - if "%CMAKE_GENERATOR%" equ "Unix Makefiles" (set "PATH=C:\cygwin\bin;%PATH%")
   - if "%CMAKE_GENERATOR%" equ "MSYS Makefiles" (set "PATH=C:\MinGW\bin;C:\MinGW\msys\1.0\bin;%PATH%")
 
-  # Use a more recent Ruby by default. We need to set the RUBY_LIBRARY variable
-  # explicitly because CMake fails to find it otherwise.
+  # Use a more recent Ruby by default.
   - set PATH=C:\Ruby21-x64;C:\Ruby21-x64\bin;%PATH%
   - ruby --version
-  - set RUBY_LIBRARY="C:\Ruby21-x64\lib\libx64-msvcrt-ruby210-static.a"
 
 before_build:
   - call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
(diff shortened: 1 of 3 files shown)


ldionne:
@brunocodutra As you can see, it's brainless but we fail to find the `RUBY_LIBRARY`. I see two options: first, we could dig deeper into the FindRuby module to find the problem, and fix it. The fix would appear in the next CMake release, I assume.

Otherwise, we could use `if (RUBY_EXECUTABLE)` in `metabench` instead of `if (RUBY_FOUND)`. Since FindRuby sets `RUBY_EXECUTABLE` regardless of whether `RUBY_LIBRARY` is found, Metabench would work even when the library (which we don't need anyway) is not found.

brunocodutra:
This issue seems to have been[fixed on February 2015](https://cmake.org/gitweb?p=cmake.git;a=commitdiff;h=e5ef9271).

Looks like we need to bump our minimum CMake version

ldionne:
Ah shit, stupid me. I actually saw this, but I thought "hey, this has been applied for a while". Turns out were were using a CMake that was released prior to that fix. So much time lost for nothing. -_-

brunocodutra:
LOL

brunocodutra:
I can't seem to find the release dates for CMake v3 series, but I infer 3.1 should ship the fix, since that's when the MSVC 14 (2015) generator (which was released late 2015) was added.

ldionne:
Ok, [I figured](https://cmake.org/Wiki/CMake_Life_Cycle_Considerations#CMake_release_dates) CMake 3.2 should have been released right after the fix.

But there's one curious thing, though. Appveyor ships with CMake 3.4, which should include the fix, yet we still get the issue. Will changing the minimum required CMake version do the trick? We'll see.

brunocodutra:
Or maybe it should be 3.2 :I

ldionne:
Well, it still doesn't work. The other option is to check for `RUBY_EXECUTABLE` instead, and it works. It would be cleaner to check for `RUBY_FOUND`, but I'm not sure digging deeper is a good time tradeoff.

brunocodutra:
> Appveyor ships with CMake 3.4, which should include the fix, 

Indeed.

Judging by the patch, it looks like FindCmake does not consider the possibility of the library name being prefixed by `lib` as it appears to be the case.

I guess we're back to checking if `RUBY_EXECUTABLE` has been set.

ldionne:
Ok, done.

PR #43, opened 2016-03-19
Provide greater control over the dependencies

brunocodutra:
The user may now:

- use locally installed versions
- disable specific libraries
- clone boost

Also, fetching libraries got much faster by downloading a zip archive instead of cloning the whole source tree (CMake does not add --depth=1)

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -35,6 +35,7 @@ if (METABENCH_BRIGAND)
             BUILD_COMMAND ""   # Disable build step
             INSTALL_COMMAND "" # Disable install step
             TEST_COMMAND ""    # Disable test step
+            UPDATE_COMMAND ""  # Disable source work-tree update
         )
         ExternalProject_Get_Property(Brigand SOURCE_DIR)
         include_directories(${SOURCE_DIR})
@@ -58,6 +59,7 @@ if (METABENCH_HANA)
             BUILD_COMMAND ""   # Disable build step
             INSTALL_COMMAND "" # Disable install step
             TEST_COMMAND ""    # Disable test step
+            UPDATE_COMMAND ""  # Disable source work-tree update
         )
         ExternalProject_Get_Property(Hana SOURCE_DIR)
         include_directories(${SOURCE_DIR}/include)
@@ -81,6 +83,7 @@ if (METABENCH_META)
             BUILD_COMMAND ""   # Disable build step
             INSTALL_COMMAND "" # Disable install step
             TEST_COMMAND ""    # Disable test step
+            UPDATE_COMMAND ""  # Disable source work-tree update
         )
         ExternalProject_Get_Property(Meta SOURCE_DIR)
         include_directories(${SOURCE_DIR}/include)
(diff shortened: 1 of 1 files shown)


benchmark/CMakeLists.txt
```diff
@@ -2,80 +2,167 @@
...
+        include_directories(${Boost_INCLUDE_DIRS})
+        add_custom_target(Boost)
+    else()
+        find_package(Git QUIET)
+        if (GIT_FOUND)
+            message(STATUS "Boost not found - fetching branch master")
+            ExternalProject_Add(Boost
+                GIT_REPOSITORY https://github.com/boostorg/boost
+                GIT_TAG origin/master
+                GIT_SUBMODULES libs/config libs/core libs/detail libs/fusion libs/mpl
```
ldionne: I'm fond of everything in your PR, but this really looks like it's going to blow up on us sometime. Did you resolve the dependencies of MPL and Fusion by hand? IMO, for Boost, it would be more reasonable and much simpler to just use `find_package`, and to forget it if it can't be found. Manually setting up Boost is a pain in the ass, and I don't think it should appear in this CMakeLists file.

benchmark/CMakeLists.txt
```diff
@@ -2,80 +2,167 @@
...
+        ExternalProject_Get_Property(Brigand SOURCE_DIR)
+        include_directories(${SOURCE_DIR})
+    endif()
+endif()
+
+if (METABENCH_HANA)
+    find_package(Hana QUIET)
+    if (Hana_FOUND)
+        message(STATUS "Hana found - version ${Hana_VERSION}")
+        #TODO: add directories here
```
ldionne: You can use `Hana_INCLUDE_DIRS` here.

benchmark/CMakeLists.txt
brunocodutra: > Did you resolve the dependencies of MPL and Fusion by hand?

Yes. A pain in the ass.

> Manually setting up Boost is a pain in the ass, and I don't think it should appear in this CMakeLists file.

That's exactly why I'd like it to be setup by the benchmark suite.

Anyways, I personally don't need that, so I can remove it, no problem.

ldionne: Another idea; instead of putting the installation of external libraries inside of this CMakeLists, perhaps we could store them in a separate file. I would care much less about installing Boost by hand in that case.

Also, we could perhaps setup a target named `benchmark-dependencies` that would perform all this setup, and remove the individual libraries from the ALL target.

ldionne: I see two resolutions:

1. We do not install Boost by hand
2. We install Boost by hand, but we install __everything__, because I don't want the dependencies thing to blow up on me two years from now, when I don't remember any of this

I'm favouring (1), because installing all of Boost every time is going to be time consuming.

brunocodutra: Agreed on 1, lets keep it simple - Boost is either installed or we disable Fusion and MPL.

benchmark/CMakeLists.txt
brunocodutra: Ok, thanks.

brunocodutra:
this should be ready to be merged

PR #45, opened 2016-03-19
fix ldionne/metabench#44

brunocodutra:
make sure metabench is able to build targets from within a directory path that contains spaces


brunocodutra:
I'll add a test for that

brunocodutra:
any ideas how we can deal with appveyor this time? ¬¬

brunocodutra:
I can't believe it, but I got it.

metabench.cmake
```diff
@@ -274,11 +274,11 @@ file(WRITE ${METABENCH_RB_PATH}
 ################################################################################
 set(MEASURE_RB_PATH ${METABENCH_DIR}/measure.rb)
 file(WRITE ${MEASURE_RB_PATH}
-"require 'benchmark'                                                    \n"
-"command = ARGV.join(' ')                                               \n"
-"time = Benchmark.realtime { `#{command}` }                             \n"
-"puts \"[command line: #{command}]\"                                    \n"
-"puts \"[compilation time: #{time}]\"                                   \n"
+"require 'benchmark'                                                         \n"
+"command = ARGV.map{ |arg| arg.match(/\\s/) ? '\"'+arg+'\"' : arg }.join(' ')\n"
```
ldionne: Nitpick: can you add a space between `map` and `{`? Also, I can't understand why this works but your previous attempt didn't; it seems that if it works quoted when there are spaces, it should work quoted when there are no spaces too?

ldionne:
Can you rebase against master and squash these 3 commits into a single one, please?

metabench.cmake
brunocodutra: > Nitpick: can you add a space between map and {? 

Sure.

> Also, I can't understand why this works but your previous attempt didn't

A side effect of quoting only when there are spaces is that we avoid quoting things like compiler options that are never expected to be quoted (I guess).

ldionne: I think something less obvious is happening. Look how the compiler command is messed up when we quote it: https://ci.appveyor.com/project/ldionne/metabench/build/1.0.185#L54

ldionne: I know this works, but it looks like something is wrong. The command line we receive in the compiler launcher is what CMake gave us; that command line should not need any modification it seems.

brunocodutra: I think there's a different issue here, probably an issue with `ctest`, because otherwise we wouldn't have any report from `cl` on STDERR, but we do

> cl : Command line error D8003 : missing source filename

brunocodutra:
Done.

PR #47, opened 2016-03-20
Automatically fetch dependencies

brunocodutra:
With this commit:
1. Dependencies (Brigand, Hana, Meta and Metal) are automatically
fetched when a dataset that depends on each of them is built
2. Benchmarks are automatically found - simply drop a folder within type
or hetero containing *.cpp.erb files and a corresponding benchmark will
be added


brunocodutra:
I really think this (or something similar) is much better than managing dozens of `CMakeLists.txt` that are almost identical.

I do understand there are situations where one might need a different range from those hardcoded by `get_range`, but I argue these aren't that common and could be handled adding a few special cases in it.

ldionne:
Closing this, since #49 was merged instead.

PR #49, opened 2016-03-20
Remove code duplication in benchmark definitions and auto fetch dependencies

ldionne:
This PR is an alternative to #47, which I find too radical. With the proposed approach, all the information essential to define a benchmark, and only that essential information, is contained in each `CMakeLists.txt` file. There is very little repetition, and I argue that this minimal amount of repetition is actually good because it gives us fine grained control over each benchmark.

This PR also fetches the dependencies of each dataset automatically, like #47.

@brunocodutra What do you think of this one?


brunocodutra:
Looks good.

ldionne:
Ok. I strongly believe that setting the ranges explicitly will pay off in the long term, but if I'm wrong we can always change it later. From my experience with benchmarking in Hana, you'll want to tweak those ranges temporarily when you try to optimize your library, and I believe you'll find it useful to be able to do that easily.

PR #50, opened 2016-03-20
Adapt which benchmarks are available based on compiler support

ldionne:
This PR is an attempt to enable/disable benchmarks based on the C++ version supported by the compiler. What's missing is how to figure out whether a compiler supports C++98, C++11 or C++14. It should be possible to find that out with CMake, or to at least approximate it using `CMAKE_CXX_COMPILE_FEATURES`.


brunocodutra:
We could use [`try_compile`](https://cmake.org/cmake/help/v3.0/command/try_compile.html) in its "first form".

The `CMakeLists.txt` we feed to `try_compile` would set `CXX_STANDARD` and `CXX_STANDARD_REQUIRED` accordingly.

ldionne:
Good suggestion. I'm trying with `try_compile`, but the first form of it.

.travis.yml
```diff
@@ -28,17 +28,17 @@ addons:
...
+  - TEST=true COMPILER=clang++-3.8 CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=clang++-3.8 CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=gcc         CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=gcc         CMAKE_GENERATOR="Unix Makefiles"
+
+  # Benchmarks
+  - BENCHMARK=true COMPILER=clang++-3.6 CMAKE_GENERATOR="Unix Makefiles"
+  - BENCHMARK=true COMPILER=clang++-3.7 CMAKE_GENERATOR="Unix Makefiles"
+  - BENCHMARK=true COMPILER=clang++-3.8 CMAKE_GENERATOR="Unix Makefiles"
+  - BENCHMARK=true COMPILER=gcc         CMAKE_GENERATOR="Unix Makefiles"
```
brunocodutra: shouldn't we do the opposite, I mean test a bunch of compilers and only benchmark on the best available (probably clang 3.8)

ldionne: My reasoning was that as long as we test on different generators and different compilers (GCC/Clang), the version of the compiler should not matter much for testing Metabench. However, it is interesting to see how a specific version of a compiler behaves in terms of compile-time, because people will be interested in seeing how the exact version they use behaves, not just the latest one.

I can add testing on all Clang versions if you think it is useful, but I would definitely benchmark on many versions.

brunocodutra: > I can add testing on all Clang versions if you think it is useful

It would be useful to test a few versions so we can check whether the detection of supported standards works.

>  I would definitely benchmark on many versions.

Makes sense of course, but I see a couple of challenges:

1. How do we push the results produced by the different Travis jobs to gh_pages?
2. Metabench would then become a tool for comparing compilers besides the original intent of comparing different implementations of metaprogramming techniques. Do we want that?

ldionne: > It would be useful to test a few versions so we can check whether the detection of supported standards works.

Ok, you're right.

> How do we push the results produced by the various Travis jobs to gh_pages?

I don't think that's an issue, we just have to push them to separate directories. For example, `clang-3.5/` gets all the benchmarks that were run on Clang 3.5, etc..


> Metabench would then become a tool for comparing compilers besides the original intent of comparing different implementations of metaprogramming techniques. Do we want that?

I do not think that would make Metabench a tool for comparing compilers (I certainly think this is out of scope for Metabench, in all cases). However, the idea is that metaprogramming libraries can have different performance profiles depending on the compiler, both because of internal compiler implementation and also special casing within a library implementation. For example, when some Clang intrinsics are available, Hana will compile faster than when they're not available. It could be the case that such intrinsics are only available starting at some version of the compiler.

All the trouble I'm going through to benchmark on multiple compilers stems from the assumption that it's useful to know how a metaprogramming library performs on a specific compiler, because compile-time performance is highly dependent on the compiler. Experience with Hana suggests that this is the case.

.travis.yml
```diff
@@ -77,13 +90,26 @@ install:
       export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${DEPS_DIR}/llvm/install/lib"
     fi
 
+  # Install Boost headers
+  - |
+    BOOST_URL="http://sourceforge.net/projects/boost/files/boost/1.60.0/boost_1_60_0.tar.gz"
+    mkdir boost && travis_retry wget --quiet -O - ${BOOST_URL} | tar --strip-components=1 -xz -C boost
```
brunocodutra: I'd rather keep it cloning from github instead, because otherwise this would require periodic maintenance in order to keep it up to date. If it takes too long, I think even manually cloning the specific submodules we need would be a better idea, even if it could break in the future.

.travis.yml
```diff
@@ -28,17 +31,27 @@ addons:
...
+  - TEST=true COMPILER=clang++-3.7 CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=clang++-3.7 CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=clang++-3.8 CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=clang++-3.8 CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=g++-4.7     CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=g++-4.7     CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=g++-4.8     CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=g++-4.8     CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=g++-4.9     CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=g++-4.9     CMAKE_GENERATOR="Unix Makefiles"
```
brunocodutra: How to deal with compilers that can't handle specific libraries due to their own bugs?
I know Hana is hard on GCC prior to version 6 and even Metal breaks GCC 4.7.

.travis.yml
brunocodutra: > I don't think that's an issue, we just have to push them to separate directories. For example, clang-3.5/ gets all the benchmarks that were run on Clang 3.5, etc..

I was more concerned about the fact jobs run in parallel, so for example it might happen that jobs A and B clone gh-pages and than both commit their changes. Then A pushes before B and so, when B tries to push, git would complain. There might be a way to tell git how to deal with it, but I wouldn't know how.

More importantly, if several jobs are pushing data to gh-pages, which job will build the front page that shows everything?

> I certainly think this is out of scope for Metabench

I agree.

> it's useful to know how a metaprogramming library performs on a specific compiler

It definitely is, but it also adds a lot of complexity. All I'm saying is we got to start thinking about it now before we bite more than we can chew.

.travis.yml
ldionne: Can we use something like `CMAKE_CXX_COMIPLE_FEATURES`?

brunocodutra: I don't think so, because these compiler versions technically do provide the features, but unfortunately it so happens that there's a bug in their implementation.

brunocodutra: I think we'll have to do some dirty logic on CMake if we want to benchmark various versions.

ldionne: Ok, you're right. Let's say for a second that we don't care about benchmarking multiple versions. Then, would you be OK with testing only on Clang 3.8 too? I'm trying to see the cost/benefit tradeoff here.

brunocodutra: Absolutely. Clang trunk would be even better, but I'm not sure where we could download binaries from.

brunocodutra: BTW I get clang trunk binaries updated daily on Archlinux through an unofficial repository, perhaps we could get that working on ubuntu. I'll look into that.

brunocodutra: [check it out](http://repos.uni-plovdiv.net/archlinux/llvm-svn/x86_64/).

ldionne: I do believe there is a clang trunk package for Ubuntu; Hana used to use it. But earlier today I couldn't find it anymore.

.travis.yml
ldionne: Ok.

.travis.yml
brunocodutra: Dumb question, did you check the history of `.travis.yaml`? :I

ldionne: I just did, and it seems like just using `clang-3.9` as an apt-package should work. But of course, clang 3.9 has not been whitelisted on Travis yet, so it won't work right now.

brunocodutra: Actually when I meant trunk, I meant whichever version that might be, not the *current* trunk.
The idea is to avoid bumping the compiler version every couple of months.

ldionne: Ok, I don't think that's possible currently.

ldionne:
This appears to be good. Still not perfect because we'd like to benchmark on Clang trunk, but at least we now manage to test and benchmark on many different compilers. As for specific compilers that do not work because of broken C++11/14 support, I see no other solution than special casing. If it gets too complicated, we may also want to give up and just support the "greatest common compiler" of all the libraries.

.travis.yml
```diff
@@ -28,17 +30,24 @@ addons:
...
+  - TEST=true COMPILER=clang++-3.6 CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=clang++-3.6 CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=clang++-3.7 CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=clang++-3.7 CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=clang++-3.8 CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=clang++-3.8 CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=g++-4.8     CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=g++-4.8     CMAKE_GENERATOR="Unix Makefiles"
+  - TEST=true COMPILER=g++-4.9     CMAKE_GENERATOR="Ninja"
+  - TEST=true COMPILER=g++-4.9     CMAKE_GENERATOR="Unix Makefiles"
```
brunocodutra: why not testing g++-5?

brunocodutra:
>  As for specific compilers that do not work because of broken C++11/14 support, I see no other solution than special casing. If it gets too complicated, we may also want to give up and just support the "greatest common compiler" of all the libraries.

Agreed.

.travis.yml
```diff
@@ -28,17 +30,24 @@ addons:
...
-  - COMPILER=clang++-3.7 CMAKE_GENERATOR="Unix Makefiles"
-  - COMPILER=clang++-3.7 CMAKE_GENERATOR="Ninja"
-
-  - COMPILER=clang++-3.8 CMAKE_GENERATOR="Unix Makefiles"
-  - COMPILER=clang++-3.8 CMAKE_GENERATOR="Ninja"
-
-  - COMPILER=gcc CMAKE_GENERATOR="Ninja"
-  - COMPILER=gcc CMAKE_GENERATOR="Unix Makefiles"
+  # Test all generators and Clang + GCC
+  - TEST=true COMPILER=clang++-3.6 CMAKE_GENERATOR="Ninja"
```
brunocodutra: Wouldn't it be worth it to test clang++-3.5, as it appears to be supported by all libraries?

ldionne: Clang 3.5 can't be installed in parallel with other versions of Clang for some reason. Since that made the CI script more complicated, and since it does not make a huge hole in the test coverage, I decided to leave it out.

.travis.yml
ldionne: I was apprehending that GCC 5 would report itself as C++14 compliant despite its support for C++14 being completely broken, which would break the benchmarks for Hana. I thought it would be best tackled in a subsequent PR.

ldionne:
I'm going to merge this knowing it's not perfect, and I suggest we fix any remaining issue in subsequent PRs. At least this will give us CI coverage for the benchmarks, which we don't have right now.

.travis.yml
brunocodutra: Isn't it because `llvm-toolchain-precise-3.5` wasn't added?

.travis.yml
brunocodutra: Ok.

.travis.yml
ldionne: No, see [this](https://travis-ci.org/ldionne/metabench/jobs/117751828#L165).

.travis.yml
ldionne: Sorry, this thread somehow went under my radar.


> More importantly, if several jobs are pushing data to gh-pages, which job will build the front page that shows everything?

No worries, Hana currently does just this to update its own benchmarks and upload them to gh-pages. Basically, each job will clone the repository and checkout the gh-pages branch. Then, each job

1. Generates its part of the benchmarks
2. Updates gh-pages (perhaps another job pushed something while this job was generating the benchmarks).
3. Commit the new changes on gh-pages
4. Push

Steps 2-4 are very quick to perform, so there is only a very small possibility that two jobs get in conflict. And even if that happens, we actually try the 2-4 steps three times before giving up and failing. So this can fail, but we'd have to be very unlucky. It has not failed yet for Hana, anyway.

> It definitely is, but it also adds a lot of complexity.

I don't think it adds too much complexity. Basically,

1. We must support all the relevant compilers in the benchmarks.
2. We must be able to pick which compiler we're displaying on the gh-pages site.

(1) is done since this PR is merged, and I don't think (2) should be complicated. But I'm cautious of not taking too large bites, and if this gets out of hand, rest assured that I'm prepared to drop this feature so we can progress, at least temporarily.

.travis.yml
brunocodutra: I see, forget it then.

.travis.yml
ldionne: In all cases, I'm not too worried about bumping the compiler version once in a while, so don't sweat too much over it. I mean, it would be nice if that could be automated, but otherwise it's just a minor annoyance.

.travis.yml
brunocodutra: > Steps 2-4 are very quick to perform, so there is only a very small possibility that two jobs get in conflict. And even if that happens, we actually try the 2-4 steps three times before giving up and failing. So this can fail, but we'd have to be very unlucky. It has not failed yet for Hana, anyway.

I see, in any case, Isn't there a way to tell git to "push merging"?

> I don't think it adds too much complexity. Basically,
> 
>    1. We must support all the relevant compilers in the benchmarks.
>    2. We must be able to pick which compiler we're displaying on the gh-pages site.

I'm actually concerned about (2), but ok, let's take a step at a time. I'd propose the following roadmap:

1. get gh-pages to display time benchmarks
2. add other aspects
3. add multiple compilers

Meanwhile, benchmarking on travis for several compilers doesn't hurt of course. Till we make it to (3) we just choose one of the compiler versions to actually push results to gh-pages.

ldionne: > I see, in any case, Isn't there a way to tell git to "push merging"?

I don't think that's possible.



> I'm actually concerned about (2), but ok, let's take a step at a time. I'd propose the following roadmap: [...]

I'm OK with your roadmap.

PR #52, opened 2016-03-20
make it possible to change the name of the HTML and JSON files

brunocodutra:
by naming them `index.{html, json}` one may take advantage of the
directory tree to display benchmarks, i.e., omitting the file name.

### DIFF
--- a/metabench.cmake
+++ b/metabench.cmake
@@ -129,7 +129,7 @@ function(metabench_add_dataset target path_to_template range)
     )
 endfunction()
 
-# metabench_add_benchmark(target [ALL] DATASETS dataset1 [dataset2 [dataset3 [...]]] [CHART path/to/json])
+# metabench_add_benchmark(target [ALL] DATASETS dataset1 [dataset2 [dataset3 [...]]] [CHART path/to/json] [FILENAME name])
 #
 #   Creates a target for running a compile-time benchmark. After issuing this
 #   command, running the target named `target` will cause each `dataset` to be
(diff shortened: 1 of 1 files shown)


ldionne:
I think the functionality is interesting, but I think `NAME` is perhaps poorly chosen to reflect that we're talking about the filename of the output files. Perhaps `FILENAME` would be better? IDK

brunocodutra:
> Perhaps FILENAME would be better?

Yeah I think so too.

metabench.cmake
```diff
@@ -147,10 +147,13 @@ endfunction()
 #       The path to a JSON file containing data for rendering the chart in the
 #       format specified by NVD3.
 #
+#   [NAME name]:
+#       A name used for the output files. Defaults to `target`.
```
ldionne: This could be made more precise by saying something like "A name used for the output files. The JSON and HTML files generated will be named ${name}.json and ${name}.html, respectively. Defaults to `target`."

ldionne:
Also, until we have figured how to publish up to date benchmarks on gh-pages, I think it would be unwise to rename the HTML and JSON benchmarks to `index`.

brunocodutra:
> lso, until we have figured how to publish up to date benchmarks on gh-pages, I think it would be unwise to rename the HTML and JSON benchmarks to index.

I just wanted to rename them to something shorter than, say, `benchmark.type.transform`, which, by the way, is also identical to their path relative to the build dir. I figured `index` would be a good choice for a name.

ldionne:
I understand, but all I'm saying is that we'll probably want to generate the benchmarks and store them on gh-pages in some way. I don't think we'll want to store the benchmarks as `xxx/index.html` and `xxx/index.json` (but I could be wrong), so I think we should figure this part out before doing any renaming.

brunocodutra:
> I don't think we'll want to store the benchmarks as `xxx/index.html` and `xxx/index.json`

In fact I think we would, because then a direct URL to a specific benchmark would be `http://xxx/benchmark/type/transform/` instead of `http://xxx/benchmark/type/transform/benchmark.type.transform.html` (I'm assuming defaulting to `index.html` would work here). But that's nitpicking indeed.

As for how we are going to present the charts on gh-pages, I've been experimenting with it locally and I already have a working example that displays all charts on a single page as a grid using `iframe`s and it works surprisingly well with NVD3. Now I just gotta get CMake to generate it automatically using ruby, but that's pretty straightforward. I should be submitting a PR soon so we can discuss about it.

brunocodutra:
Just to make things clear, I'm not opposing to neither changing the option to `FILENAME` nor to postpone the decision regarding what name the files should have, in fact I'll commit these changes ASAP, I just can't do that now.

ldionne:
No worries, I'm not in a hurry.

> As for how we are going to present the charts on gh-pages, I've been experimenting with it locally and I already have a working example that displays all charts on a single page

Awesome!

brunocodutra:
Done.

PR #54, opened 2016-03-21
fix bug which caused _dependency to not be unset when a library was disabled

brunocodutra:
- rename _supported to _enabled so as to make its meaning clearer


brunocodutra:
Resolved merge conflict.

PR #55, opened 2016-03-21
increase timeout for download operations

brunocodutra:
according to CMake documentation

> [TIMEOUT seconds] # Time allowed for file download operations

10 seconds was far too short

**NOTE:** this is based on #54


brunocodutra:
Resolved merge conflict.

PR #58, opened 2016-03-22
add an index.html that displays all charts

brunocodutra:
target `benchmarks` now generates an `index.html` within its binary
directory that displays all charts in a grid.


brunocodutra:
This is an early working draft of what should become `gh-pages`.
 
A few things still need to be addressed, including:

1. How should heterogeneous and pure type benchmarks be segregated?
2. How should we display executable size benchmarks?
3. Should we provide an index for quick reference? How?
4. Anything else?

Regarding 1., how about providing two *tabs* on the top navbar, one for type and another for hetero?

Regarding 2. I was thinking we could split the grid vertically and display time benchmarks side-by-side with the corresponding size benchmark, what do you think?

Regarding 3. I think an index is a must once we start growing the number of benchmarks. How about providing a third *tab* named Alphabetical Index inspired by [this one](http://brunocodutra.github.io/metal/namespacemembers.html)?

ldionne:
> This is an early working draft of what should become gh-pages.

I __love__ it so far! Thanks for your work!


> A few things still need to be addressed, including:
> 
> 1. How should heterogeneous and pure type benchmarks be segregated?
> 2. How should we display executable size benchmarks?
> 3. Should we provide an index for quick reference? How?
> 4. Anything else?
>
> Regarding 1. how about providing two tabs on the top navbar, one for type 
> and another for hetero?

Yes, I believe that would work. I guess we have to try it to see how it looks, but I think pure type-level computations must be seggregated from heterogeneous computations, since they are different things.


> Regarding 2. I was thinking we could split the grid vertically and display 
> time benchmarks side-by-side with the corresponding size benchmark, what do 
> you think?

I think it might be more relevant to display RAM usage side by side with compile times. Also, for type-level computations, the executable size is mostly irrelevant so I wouldn't do that. For heterogeneous computations, however, both RAM usage and executable size would be valuable. I think there's a challenge in how to present this information in a meaningful way.

> Regarding 3. I think an index is a must once we start growing the number of benchmarks. How about providing a third tab named Alphabetical Index inspired by this one?

Actually, what if the charts were all "minimized" with only their titles showing, and when you click on the chart that would expand to a box showing the chart itself. That would achieve the same function as an index, but without adding an additional page. It should also be possible to expand all the charts to get a grid similar to what we have right now.

Also, the chart titles are not showing right now.

brunocodutra:
> > This is an early working draft of what should become gh-pages.
> 
> I love it so far! Thanks for your work!

Awesome :D

> I think it might be more relevant to display RAM usage side by side with compile times. Also, for type-level computations, the executable size is mostly irrelevant so I wouldn't do that. For heterogeneous computations, however, both RAM usage and executable size would be valuable. I think there's a challenge in how to present this information in a meaningful way.

You're right, this is a tricky one, let's leave it on hold for now then.

> Actually, what if the charts were all "minimized" with only their titles showing, and when you click on the chart that would expand to a box showing the chart itself.

Great idea!

> Also, the chart titles are not showing right now.

That's because we don't have a `chart.json` per benchmark yet, BUT I'm growing fond of the idea of getting rid of them and, instead, do all the chart configuration via CMake, either via `metabench_add_benchmark` or a third function, say `metabench_add_chart`. The latter would have the benefit of allowing us to configure various charts per benchmark individually (think one per aspect), but that would also be trickier to implement.

ldionne:
> BUT I'm growing fond of the idea of getting rid of them and, instead, do all the chart configuration via CMake, either via metabench_add_benchmark or a third function, say metabench_add_chart

Actually, if we go this way, I think `metabench_add_benchmark` would become completely unnecessary. In my view, we really have two different things. One thing is datasets, which represent a collection of data related to a C++ file, and the other one is charts, which are just an aggregation of several datasets into a visual form. But in all cases, I suggest we get the gh-pages thing working before we do any conceptual change (even small) to the CMake module itself.

brunocodutra:
> Actually, if we go this way, I thinkmetabench_add_benchmark would become
completely unnecessary.

Not really, a `benchmark` represents data and thus produces a JSON file
that contains the results from various datasets. A `benchmark` can then be
consumed by several `charts` that represent views of that data, each for an
aspect for example.

Notice the JSON file produced by a benchmark contains data for all aspects
and therefore is not tied to a specific view, as a chart would be.

>  I suggest we get the gh-pages thing working before we do any conceptual
change (even small) to the CMake module itself.

We can do that, but notice that we are going to need this abstraction to
get that "collapsible panel thing" going, since we gotta get the chart's
title from somewhere and that would be from the target added by
`metabench_add_chart`.

ldionne:
Well, if we offload the chart generation to some `metabench_add_chart`, then `metabench_add_benchmark` becomes nothing more than an aggregation of multiple datasets into a single JSON file, and I question the value of this. If we keep `add_benchmark` and add `add_chart`, then we have

```
dataset1, ..., datasetN -> benchmark -> chart1, ..., chartN
```

What I propose is to simply skip the (IMO) artificial step of aggregating many datasets into a single JSON file, and to simply have the `add_chart` fetch the datasets directly:

```
dataset1, ..., datasetN -> chart1, ..., chartN
```

In other words, what I propose is that `add_chart == add_benchmark` as currently defined. As for getting the gh-pages thing going, can't we just use whatever title is provided in `chart.json` as the title? We would simply need to add proper `chart.json` files to the benchmarks we have (temporarily). It seems like this would be the least risky way to get gh-pages going ASAP.

Then, right after we get it going, I would suggest we

1. Add a `TITLE` argument to `metabench_add_benchmark`, which could default to the benchmark's target.
2. Remove the (now unneeded) `chart.json` files from the `benchmark/` folder.

It seems to me like this path is the most straightforward towards getting gh-pages to work, but let me know if you think I'm missing something.

brunocodutra:
> Well, if we offload the chart generation to some `metabench_add_chart`, then `metabench_add_benchmark` becomes nothing more than an aggregation of multiple datasets into a single JSON file, and I question the value of this.

The concept of `benchmark` would still be necessary in order to generate the JSON. If we were to not generate it any more, then I agree with you that the concept of `benchmark` would be equivalent to the concept of `chart`. 

Alright, so I see you don't actually need the JSON file that we currently generate, and nor do I, so I propose we get rid of it and, instead, make the JSON files produced by individual datasets public, i.e. move them out of `_metabench` and properly document it. If we do that, then I would agree with you that a `benchmark` would represent a view over a set of `datasets` and I would also agree with everything else you said.

ldionne:
> Alright, so I see you don't actually need the JSON file that we currently generate, and nor do I, so I propose we get rid of it

The JSON file that we generate is useful to me right now (I'm using Metabench for upcoming talks), but I'll see what it would mean to use the datasets directly instead. It should have hit me sooner, but now I see the value of having a separate `add_benchmark` and `add_chart`.

> make the JSON files produced by individual datasets public, i.e. move them out of _metabench and properly document it.

Sorry; I should have said it, but my proposal included this change too.


I'll try to understand exactly what I need better, and then we can discuss on the add_benchmark vs add_chart issue. For now, I suggest the status quo and we just use `chart.json` to achieve the gh-pages stuff. I'll create a separate issue for the add_chart vs add_benchmark.

brunocodutra:
> For now, I suggest the status quo and we just use chart.json to achieve the gh-pages stuff. I'll create a separate issue for the add_chart vs add_benchmark.

Ok, agreed.

ldionne:
Here's what I'm thinking about. The Travis script(s) would update gh-pages by generating their benchmarks on their assigned compiler, and then they would push this to gh-pages in a subdirectory named `benchmarks/${COMPILER}`.

So basically, gh-pages would look like

```
gh-pages/
    index.html
    benchmarks/
        clang++-3.6/
            benchmark1.json
            ...
            benchmarkN.json
        clang++-3.7/
            ...
```

For a start, `index.html` would simply need to know about all of the benchmarks (which it already does), and we'd pick an arbitrary compiler. Does that make sense?

ldionne:
Forget the above, I got carried away. Now I see that just running `make benchmarks` will generate both the benchmarks and the index.html, and we'll just have to copy the `index.html` to gh-pages. This is indeed the simplest way to setup things for a start.

ldionne:
Ok, I really love the look of this so far. Once the CI passes (and it will), I'll merge this so we can both work on it (if that's OK with you).

brunocodutra:
> Now I see that just running make benchmarks will generate both the benchmarks and the index.html, and we'll just have to copy the index.html to gh-pages.

Beware, besides the `index.html` we need to push `*.js`, `*.css`, `*.eot` and `*.ttf`, as well as every `*.html` within `type/` and `hetero/`, preserving their relative paths.

ldionne:
Yes, I figured out. Sorry for the above confusion :). Can I merge?

brunocodutra:
> Once the CI passes (and it will), I'll merge this so we can both work on it (if that's OK with you).

Yeah of course.

brunocodutra:
Don't forget to add a link to the `gh-pages` on the front page.

brunocodutra:
Also, regarding multiple compilers, if we push results to `<compiler>/{type, hetero}`, we may simply add a drop-down menu on the top navbar listing the available compilers. Depending on what's selected, we point the `iframe`s to the correct prefix.

Quite simple.

ldionne:
Yup, that's what I thought we'd do. I'm currently adding support for pushing all the benchmarks to the gh-pages branch from Travis (see #59).

ldionne:
> Beware, besides the index.html we need to push *.js, *.css, *.eot and *.ttf, as well as every *.html within type/ and hetero/, preserving their relative paths.

Actually, regarding the `.js`, `.css` and other files, do you see a reason not to put them directly on the gh-pages branch and leave them there? Then, we'd only have to update `index.html` and the benchmarks.

brunocodutra:
> do you see a reason not to put them directly on the gh-pages branch and
leave them there?

Not at all, in fact I'd be glad to get rid of `benchmark/web/` on master.

ldionne:
Alright, I'll do that. BTW, I'm currently looking at the possibility of using the simplest Jekyll setup ever to generate the website. We would only have to update the benchmark data. Just a WIP at the moment, though.

brunocodutra:
> I'm currently looking at the possibility of using the simplest Jekyll
setup ever to generate the website.

I'm not familiar with Jekyll, sounds fancy, what is the idea?

ldionne:
It's a static site generator supported by GitHub pages. Basically, we have an `index.html` file with some special markup in it (similar to ERB markup), and we just push that to gh-pages. Then, GitHub Pages will process that markup to generate an actual html file that we can view. It's used a lot to generate static blogs, but our use case is so simple that I think it'll work almost perfectly.

brunocodutra:
> Basically, we have an index.html file with some special markup in it (similar to ERB markup), and we just push that to gh-pages.

Cool! So we just gotta port our brainless ERB to Jekyll?

ldionne:
Yes, basically. I'm looking into it right now.

PR #63, opened 2016-03-23
Use Jekyll to generate the static site

ldionne:
This, in combination with [these](https://github.com/ldionne/metabench/compare/2430d76~...97e7a40d) changes on gh-pages, are what's required to use Jekyll on gh-pages. This allows us to put all the website-related stuff directly on the gh-pages branch, which I think is good.

Let me know what you think. If there's a good reason not to move forward with this, I'll roll back the changes I made to gh-pages.


brunocodutra:
Looks very good. 
The only thing I'm concerned about, is being able to test locally before merging to master, but I guess I could setup gh-pages on my fork for that.

brunocodutra:
Also, URLs to specific charts look awkward now, for instance `metabench/hetero/benchmark.hetero.at.html`

I'd prefer something like `metabench/hetero/at.html` or `metabench/hetero/at` (defaults to `index.html`). Remember that URLs are public and I for one would like to be able to link specific charts.

To achieve the latter option, for example, it would just be a matter of reverting FILENAME to `index` and amend travis.yml to copy files preserving their relative path (there must be a flag for that).

I would also use the opportunity to already place charts within a directory named after the compiler (cang-3.8 in this case), since that's how it's gonna be in the end anyways, so links wouldn't change any more from now on.

ldionne:
> The only thing I'm concerned about, is being able to test locally before merging to master, but I guess I could setup gh-pages on my fork for that.

Checkout the `gh-pages` branch and then use `jekyll serve --baseurl ''`.

ldionne:
> Also, URLs to specific charts look awkward now, for instance metabench/hetero/benchmark.hetero.at.html

I agree the URLs look strange. Noted in #64.

> I would also use the opportunity to already place charts within a directory named after the compiler (cang-3.8 in this case)

That is fixed by #59.

Issue #69, opened 2016-03-24
[gh-pages] Should we smooth benchmark results?


brunocodutra:
Right now charts are displayed exactly as measured and are thus subject to variations related to thread scheduling and other SO affairs that result in undesirable spikes. 

Activating [`basis` or `bundle` interpolation on NVD3](http://www.d3noob.org/2013/01/smoothing-out-lines-in-d3js.html) should smooth out most spikes.

ldionne:
I guess we can try it out, but I would be tempted to think that we should show raw data and document how the benchmarks are gathered (on Travis CI instances, etc...). This way, people can apply some kind of filter on the results they see.

If it makes the charts much easier to visualize, however, then it might be worth it.

brunocodutra:
Raw data will always be available through the JSON files and also even in the source file of the HTML files, as data manipulation would be carried out by javascript.

ldionne:
Ok, then let's try it out.

ldionne:
BTW, should we publish the JSON files too? I think it makes sense to do so.

brunocodutra:
Yes, ideally we should provide some public API that others could use load benchmark data dynamically into their projects, something in the lines of [GH's API](https://developer.github.com/v3/), but I have no idea how to accomplish that or even if it is possible at all in our current setup.

brunocodutra:
I've been experimenting with interpolation locally and while it improves the situation, spikes are still quite noticeable.

Compare raw values
![raw](https://cloud.githubusercontent.com/assets/4043663/14082535/e72a0b7a-f4e6-11e5-9567-b30706a3fe34.png)

with D3 `basis` interpolation
![interpolated](https://cloud.githubusercontent.com/assets/4043663/14082528/e13dfa50-f4e6-11e5-9314-bb303f1d8c8d.png)

What do you think?

brunocodutra:
BTW, what could be the cause of the severe spikes in the benchmarks of Fusion and MPL? I suspect the preprocessor could be the culprit? Would there be any way to  mitigate this? 

How about we repeat each compilation step a couple of times (3) and take the median value as the actual benchmark? I understand benchmarking would take a couple of times (3) longer, but do we care? Are we near blowing up Travis limits? 

Perhaps it would be a good idea to reduce the number of samples and increase the number of times we repeat each step.

ldionne:
> Compare raw values 
> [...]
>
> with D3 basis interpolation
> [...]
> 
> What do you think?

[...]

PR #74, opened 2016-03-29
Use the median-of-3 when gathering timings

ldionne:
With this PR, the timings are done 3 times for each `n` in the range, and the median of the 3 timings is taken. This is an experiment to check whether the timings appear to be more reliable this way, as discussed in #69. Note that running the benchmarks now takes 3 times as long, which could be a deal breaker.


brunocodutra:
Looks good, did it improve the reproducibility of benchmark results?

ldionne:
> Looks good, did it improve the reproducibility of benchmark results?

Haven't checked yet. I'm trying to think of a way to compare reproducibility, since just running the benchmarks with and without the median trick won't tell us much. I think this would work: we could generate the benchmarks multiple times without the median trick, and then multiple times with the median trick. Then, we would look at the average dispersion of the curves with and without the trick. Does that make sense?

brunocodutra:
> We could generate the benchmarks multiple times without the median trick, and then multiple times with the median trick. Then, we would look at the average dispersion of the curves with and without the trick. Does that make sense?

That is exactly what I mean.

Also, the trick could be made into a feature, i.e. `metabench_add_dataset` could take an additional parameter that defaults to `1` that sets the number of times each compilation should be repeated. We could even decide to enable the trick for MPL and Fusion only and perhaps even set a higher number of repetitions for them (specially for Fusion).

metabench.cmake
```diff
@@ -247,6 +247,11 @@ file(WRITE ${METABENCH_RB_PATH}
 "require 'time'                                                                                         \n"
 "                                                                                                       \n"
 "module Metabench                                                                                       \n"
+"  def self.median(values)                                                                              \n"
+"    values = values.sort                                                                               \n"
+"    return values[values.length / 2]                                                                   \n"
```
brunocodutra: When `values.length` is even, the median should be `(values[values.length / 2 - 1] + values[values.length / 2])/2`

ldionne:
Adding a customization point adds complexity, and I'm really not sure of the added value. Also, if we go this way, I don't see a reason not to allow deciding whether we pick the median or the mean of the n observations, which would add even more complexity. I think we must be careful not to over-engineer the data acquisition, since this is a very deep hole.

metabench.cmake
ldionne: Fixed.

brunocodutra:
> think we must be careful not to over-engineer the data acquisition, since this is a very deep hole.
View changes

I agree, but in this case it could be useful to reduce the time required for running the benchmarks and thus avoid blowing up Travis limits. Anyway, lets first assess whether it improves reproducibility.

ldionne:
At first glance, it's a bit difficult to tell whether the median trick really gives an improvement. I run the benchmarks 11 times with and without the median trick, and here are the generated results.

The 11 runs with the median trick:  [median.zip](https://github.com/ldionne/metabench/files/198829/median.zip)
The 11 runs without the median trick: [no_median.zip](https://github.com/ldionne/metabench/files/198830/no_median.zip)

Now, I might have been using my computer during some of these benchmarks (nothing intensive, though), but then again the benchmarks are run on Travis and the cores can be shared, so we should be able to deal with that.

I think the general trend is that the median trick does improve the results, in the sense that we rarely get curves that are completely foobar.

ldionne:
If we do provide this feature, it should be optional because that'll blow up the benchmark times, and that's not acceptable.

ldionne:
Also note that the difference seems to be much greater for benchmarks of heterogeneous computations than for pure type computations.

ldionne:
So the bottom line is that it's slightly better (it couldn't worsen things, since it's a median), but the benefit may not be so huge. Do we want to make the data acquisition for this benefit? Maybe, I'm not sure. I'll wait for your input.

brunocodutra:
>  run the benchmarks 11 times with and without the median trick, and here are the generated results.

Now that's what I call professionalism.

> I'll wait for your input.

In fact I think the improvement is quite noticeable and I'm sure it would get increasingly better the larger the size of the median filter, so I think it is definitely worth it to provide the feature even if we might not be able to take full advantage of it for the benchmark suite due to Travis constraints. Remember that others will hopefully find Metabench useful for their needs and a median filter might come in handy.

In any case it should be quite easy to provide the feature from this point, so why not?

ldionne:
I would say this fixes #69.

brunocodutra:
Looks great.

Issue #60, opened 2016-03-22
[Design] Charts vs benchmarks


ldionne:
Right now, we have `metabench_add_dataset` and `metabench_add_benchmark`. It was suggested in #58 that we might want to add `metabench_add_chart`. Let's use this thread for further discussion on this issue.

brunocodutra:
I believe this has been addressed by #101.
It was decided that `metabench_add_benchmark` should be understood as semantically equivalent to adding a chart, thus a separate `metabench_add_chart` isn't required.

ldionne:
Actually, I was thinking that we should change the name to `metabench_add_chart`. What do you think?

brunocodutra:
> Actually, I was thinking that we should change the name to
metabench_add_chart. What do you think?

I think it would be better indeed

PR #81, opened 2016-03-31
Add default axis labels

brunocodutra:
Right now we only plot time, so we might as well display proper axis labels.


metabench.cmake
```diff
@@ -377,6 +377,8 @@ file(WRITE ${CHART_HTML_ERB_PATH}
 "        /* configure the x axis */                                          \n"
 "        var xAxis = chart.xAxis || {};                                      \n"
 "        delete(chart.xAxis);                                                \n"
+"        if(typeof xAxis.axisLabel == 'undefined')                           \n"
+"          xAxis.axisLabel = 'Number of Elements'                            \n"
```
ldionne: The X axis is not always a number of elements. For example, for the `at` benchmark, it's actually the index of the element we're accessing. (Granted, it's also the number of elements in the sequence for now but that will change since this benchmark is broken as is).

ldionne: I think the best we can say is `n`, and then we must explain what `n` is on a per-chart basis.

brunocodutra: > The X axis is not always a number of elements.

Well, this is just the default, we can still configure it to something else, which brings me back to #60. I think it's time we settle it before we add a bunch of unnecessary `chart.json`.

> this benchmark is broken as is

I wouldn't call it broken, but we really should rename it to `back`. BTW, now that the gh-pages is up and running, what about expanding our benchmark pool?

brunocodutra:
Rebased onto master.
Anything preventing this PR to be merged?

ldionne:
> Anything preventing this PR to be merged?

Oversight. I also think it would be a good time to expand our benchmark pool; I'll try to find some time to do it this week.

brunocodutra:
> I also think it would be a good time to expand our benchmark pool; I'll try to find some time to do it this week.

Awesome.

PR #88, opened 2016-04-02
add a direct link for each chart

brunocodutra:


brunocodutra:
Do you think you could manually merge this? I don't know why git complains, it should simply be a matter of adding these lines.

ldionne:
I was able to rebase it onto `gh-pages`; it fails to merge automatically because the surrounding context changed. However, this adds the button to the bar itself, and thus clicking on the button causes the chart to expand, which I think is odd.

Try ctrl+click on the button; it will open the chart in a new tab __and__ expand the chart. I think it should only open the chart in a new tab.

brunocodutra:
> However, this adds the button to the bar itself, and thus clicking on the button causes the chart to expand, which I think is odd.

On what browser are you testing? It didn't behave like this on chromium at least.

ldionne:
Safari, but Chrome behaves the same. I must have changed something in the original PR when rebasing it.

brunocodutra:
> Safari, but Chrome behaves the same. 

Weird, I'll investigate.
In any case stopping propagation should fix this issue, thanks for testing.

brunocodutra:
Rebased atop `gh-pages`.

I still can't reproduce the issue on neither:

* Chromium 49.0.2623.110
* Mozilla Firefox 45.0.1

Could you test again please? Specially on Safari to which I don't have access.

ldionne:
metabench-panel-expansion

First is Safari. You can see that each time I cmd+click the little icon, a new tab opens up with the chart, but it also expands the panel.

brunocodutra:
I'll be damned.

brunocodutra:
Oh, now I get it, ctrl + click is different from clicking the scroll button.

ldionne:
IDK, my trackpad has no scroll button :p

brunocodutra:
should be fixed now

Issue #89, opened 2016-04-02
The benchmark suite fails to compile using C++98


brunocodutra:
Even though Fusion was developed in the C++03 era, its benchmarks make use of C++11 features such as `constexpr` and `auto`.

PR #90, opened 2016-04-03
make sure the benchmark suite compiles under C++98

brunocodutra:
Fixes ldionne/metabench#89

### DIFF
--- a/benchmark/FindCXXStandard.cmake
+++ b/benchmark/FindCXXStandard.cmake
@@ -16,34 +16,42 @@ file(WRITE "${SOURCE_FILE}"
     "int main() { }                                                 \n"
 )
 
-# Check for C++14
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+message(STATUS "Check for C++14")
 try_compile(_supports_14 ${CMAKE_BINARY_DIR} ${SOURCE_FILE}
     CMAKE_FLAGS -DCMAKE_CXX_STANDARD=14
     COMPILE_DEFINITIONS -DSTANDARD_DATE=201402L
 )
 if (_supports_14)
     set(CXX_STANDARD 2014)
+    set(CMAKE_CXX_STANDARD 14)
+    message(STATUS "Check for C++14 -- works")
     return()
 endif()
 
-# Check for C++11
+message(STATUS "Check for C++11")
 try_compile(_supports_11 ${CMAKE_BINARY_DIR} ${SOURCE_FILE}
     CMAKE_FLAGS -DCMAKE_CXX_STANDARD=11
     COMPILE_DEFINITIONS -DSTANDARD_DATE=201103L
 )
 if (_supports_11)
     set(CXX_STANDARD 2011)
+    set(CMAKE_CXX_STANDARD 11)
+    message(STATUS "Check for C++11 -- works")
     return()
 endif()
 
-# Check for C++98
+message(STATUS "Check for C++98")
 try_compile(_supports_98 ${CMAKE_BINARY_DIR} ${SOURCE_FILE}
     CMAKE_FLAGS -DCMAKE_CXX_STANDARD=98
     COMPILE_DEFINITIONS -DSTANDARD_DATE=199711L
 )
 if (_supports_98)
     set(CXX_STANDARD 1998)
-    return()
+    set(CMAKE_CXX_STANDARD 98)
+    message(STATUS "Check for C++98 -- works")
+return()
 endif()
 
 # Emit an error if none of the standards are satisfied.
--- a/benchmark/hetero/at/CMakeLists.txt
+++ b/benchmark/hetero/at/CMakeLists.txt
@@ -2,9 +2,6 @@
 # Distributed under the Boost Software License, Version 1.0.
 # (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 set(datasets)
 add_dataset(datasets dataset.hetero.at.fusion.list      fusion.list.cpp.erb         "(1..50).step(2)")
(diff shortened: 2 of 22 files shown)


benchmark/FindCXXStandard.cmake
```diff
@@ -16,38 +16,32 @@ file(WRITE "${SOURCE_FILE}"
...
-endif()
-
-# Emit an error if none of the standards are satisfied.
-message(FATAL_ERROR
-    "FindCXXStandard.cmake: The C++ compiler '${CMAKE_CXX_COMPILER}' appears "
-    "not to support any of the known standard (1998, 2011, 2014)."
-)
+message(STATUS "Fall back to C++98")
+set(CXX_STANDARD 1998)
+set(CMAKE_CXX_STANDARD 98)
```
brunocodutra: Testing for C++98 support wouldn't hurt, if it weren't for buggy compilers such as GCC 4.6 which sets `__cplusplus` to `1L`, but since we set `CMAKE_CXX_STANDARD_REQUIRED`, it's better we let CMake error out, should someone try running the benchmark suite for 20 years old compilers which indeed don't support C++98.

brunocodutra: Also note that CMake deems GCC 4.6 compliant to C++11, since it provides `-std=c++0x`, however it fails our test becuase it sets `__cplusplus` to 1L, so it ends up falling back to C++98. This allow us to test the benchmark suite under C++98, but if you look at the Travis log, you'll realize it successfully compiles the C++11 examples, even though only Fusion and MPL benchmarks are run.

ldionne:
General comment: since one of the primary goals of the benchmark suite is to compare C++03 libraries with C++11 libraries, I would expect C++11 to always be available. I'm not sure of the value added by accommodating older compilers.

Regarding GCC 4.6, it's technically not a _compliant_ C++11 compiler, since `__cplusplus` is not defined properly. My idea when writing this was that we don't care for older compilers such as GCC 4.6, but maybe I'm wrong and we do care about them.

brunocodutra:
> General comment: since one of the primary goals of the benchmark suite is to compare C++03 libraries with C++11 libraries, I would expect C++11 to always be available. I'm not sure of the value added by accommodating older compilers.

Well, since we test for the compliance to standards and already have all the machinery required to enable benchmarks based on compiler support I felt it didn't make much sense to require C++11 for MPL and fusion. (Actually we used to require C++14 for all benchmarks through CMake).

> Regarding GCC 4.6, it's technically not a compliant C++11 compiler, since __cplusplus is not defined properly. My idea when writing this was that we don't care for older compilers such as GCC 4.6, but maybe I'm wrong and we do care about them.

Sure, the idea behind this PR is not to being able to benchmark GCC 4.6, but doing so is the easiest way to test whether our Fusion and MPL benchmarks comply to C++98.

benchmark/FindCXXStandard.cmake
```diff
@@ -16,34 +16,42 @@ file(WRITE "${SOURCE_FILE}"
...
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+message(STATUS "Check for C++14")
 try_compile(_supports_14 ${CMAKE_BINARY_DIR} ${SOURCE_FILE}
     CMAKE_FLAGS -DCMAKE_CXX_STANDARD=14
     COMPILE_DEFINITIONS -DSTANDARD_DATE=201402L
 )
 if (_supports_14)
     set(CXX_STANDARD 2014)
+    set(CMAKE_CXX_STANDARD 14)
```
ldionne: I'd rather keep setting `CXX_STANDARD`. We don't want to overwrite `CMAKE_CXX_STANDARD`, which is used for something different.

benchmark/FindCXXStandard.cmake
```diff
@@ -16,34 +16,42 @@ file(WRITE "${SOURCE_FILE}"
     "int main() { }                                                 \n"
 )
 
-# Check for C++14
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+message(STATUS "Check for C++14")
```
ldionne: If we want to print a message, let it be something like `"C++ standard found to be ..."`, and print it only once to avoid cluttering the CMake generation step. We just have to put the message inside the `if(_supports_xx)` statement.

benchmark/FindCXXStandard.cmake
ldionne: Oh wait, I missed what you were doing. Forget this comment.

benchmark/FindCXXStandard.cmake
ldionne: This is not a big deal; I'll merge to clear up some PRs.

brunocodutra: The idea was just to have CMake report which standard it detected.
I'm ok either way.

brunocodutra:
Just a heads up: I reverted the commit which added GCC 4.6 to test compliance to C++98, because it actually relied on a bug and would certainly break in the future, so currently we don't have any way of testing whether MPL and Fusion benchmarks still comply to C++98.

In fact I don't think there's an easy way besides enforcing C++98 via `CMAKE_CXX_STANDARD`, since I don't think travis provides a compiler without at least experimental C++11 support and our current setup just picks the greatest supported feature set.

ldionne:
I don't see this as a problem. For the benchmark suite, the only compilers we care about are those that we run the benchmarks with. As I've said before, I don't even see the usage of C++11/14 features when C++98 would be sufficient as a problem, as long as the benchmark suite works with the compilers that we're interested in.

Basically, I would settle on a list of compilers that we care about, and then make sure the benchmark suite compiles under all these compilers (when reasonable, e.g. Hana won't compile under MSVC). As long as the benchmark suite works, my first concern would be making the benchmarks as similar as possible for different libraries much more than making them minimal w.r.t. modern C++ usage.

brunocodutra:
> I don't see this as a problem

Me neither, it was just a heads up since I changed the PR without notice. If it were "free" to have it tested I'd be in favour of adding it, but it isn't. 

Also, this PR was more a side effect of fixing some inconsistenses I found regarding the selection of C++ standard, no big deal.

Issue #84, opened 2016-04-01
[gh-pages] Provide benchmarks on MSVC and GCC 5


brunocodutra:
We will have to deal with broken support for newer standards. Two approaches strike me as the most obvious:

* Manually implement some nasty logic to decide what libraries to enable for each compiler version.
* Try compiling each benchmark (rendered for `n=1`) before deciding whether the selected compiler version is able to handle a library.

brunocodutra:
I'm fond of the second approach, because it requires the least maintenance and because it looks cool.

brunocodutra:
> I'm fond of the second approach, because it requires the least maintenance and because it looks cool.

Forget it, too complex to be worth it.

ldionne:
Sorry, I was out for the weekend. The issue with the second approach is that we won't know whether it's a compiler failure or a problem in the benchmark itself. IOW, we wouldn't be able to "test" that the benchmarks are not broken.

brunocodutra:
>  Sorry, I was out for the weekend.

No worries.

Do you think you can setup Appveyor to push benchmark results in a manner similar to Travis?

brunocodutra:
Also, just realized we never tested MSVC, as the Ninja generator picks MinGW.
Working on it right now.

brunocodutra:
Surprise surprise, MSVC sets `__cplusplus` to `199711L`.

Ideas?

ldionne:
That's because MSVC is not C++11/14 compliant AFAIK.

ldionne:
So there's nothing to do here; any benchmark that actually require full C++11/14 compliance can't support MSVC. Now, we could go with more granularity and require specific features `cxx_constexpr`, etc.. using `COMPILER_FEATURES` in CMake, but that's __really__ not worth the added complexity, I think.

brunocodutra:
> That's because MSVC is not C++11/14 compliant AFAIK.

It doesn't, but most libraries we actually care about don't need it either. Also see #93.

brunocodutra:
As a [matter of fact](https://ci.appveyor.com/project/brunocodutra/metabench), Metal works from MSVC 14 onwards and Brigand from MSVC 12 onwards, not to mention MPL and Fusion which date back to over a decade ago.

[...]

PR #92, opened 2016-04-03
Skip compiler versions not supported by specific libraries

brunocodutra:
This PR allows benchmarking GCC 5 (#84)

**Note:** This is based on #90


.appveyor.yml
```diff
@@ -14,14 +14,17 @@ build:
 
 environment:
     matrix:
-      - CMAKE_GENERATOR: "Ninja"
-      - CMAKE_GENERATOR: "MinGW Makefiles"
```
ldionne: Are we not testing with MinGW Makefiles anymore?

brunocodutra: Not because I didn't want to, I simply couldn't get it to work. If MinGW is in the path, then Ninja gets messed up in the head and tries linking MinGW libraries to objects compiled by MSVC, then If I try to conditionaly remove it from `PATH`, then `cmd` gets high on acid and starts complaining about some "google cloud" thingy - somehow setting PATH within an `if` statement doesn't work on appveyor, I tried quoting the hell out of it, but to no avail.

Since MinGW is essentially GCC on windows, I felt there was no added value in testing it.

brunocodutra: Also, I think this is the first time we are actually testing MSVC, because Ninja would always pick MinGW and I never noticed it.

brunocodutra:
I'll update this PR to get rid of `FindCXXStandard` in the lines of #94.

.appveyor.yml
ldionne: Ok, then I'm totally OK with dropping MinGW.

Issue #93, opened 2016-04-04
MSVC 14 still doesn't report compliance to C++11


brunocodutra:
Using `__cplusplus` to decide what standard is supported is too restrictive.
A more flexible alternative is to use [CMake compiler features](https://cmake.org/cmake/help/v3.1/manual/cmake-compile-features.7.html).

brunocodutra:
Related to #84 #89 #90 #92

brunocodutra:
I think I found a very simple way to fix `FindCXXStandard`, the idea is to `try_compile` a bogus cpp file while setting `CMAKE_CXX_STANDARD` and also `CMAKE_CXX_STANDARD_REQUIRED`.

If this works as intended, then `FindCXXStandard` will report the exact CXX standard CMake knows to be supported by the compiler, which is perfect.

brunocodutra:
>  the idea is to try_compile a bogus cpp file while setting CMAKE_CXX_STANDARD and also CMAKE_CXX_STANDARD_REQUIRED.

Unfortunately this doesn't work because CMake doesn't recover from a `FATAL_ERROR` within `try_compile`:

`CMake Error: Internal CMake error, TryCompile generation of cmake failed`

brunocodutra:
I've been thinking about it and perhaps we should forget about detecting the "highest supported C++ standard" in favour of version checking, as this is a rather artificial concept in a world just about two compilers *actually try* to thoroughly  comply with every aspect of a standard revision. Moreover, while looking at [`CMAKE_CXX_KNOWN_FEATURES`](https://cmake.org/cmake/help/v3.1/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES), I came to realize Metal for one needs just about half a dozen of the many dozens listed there and I'm sure it isn't different for the other libraries we currently care about.

In addition to compiler version checking, we could also write a CMake test for each library that detects the features it requires (we could even use Boost.Config to broaden compiler support), but I doubt the value of this, since we will have to do version checking one way or the other to deal with buggy compilers.

brunocodutra:
Closing this in favour of #84

PR #94, opened 2016-04-04
Use CMAKE_CXX_COMPILE_FEATURES to enable/disable libraries

ldionne:
This PR is an attempt to enable/disable libraries based on the features supported by the compiler instead of the official supported C++ version, which is too coarse grained.


brunocodutra:
Following my arguments on #93, while explicitly checking for known features makes perfect sense, I doubt the value of it given the fact it won't avoid explicit checking compiler versions to work around compiler bugs. 

I propose we take the pragmatic path and forget about trying to detect what compilers are supported and simply add exceptions based on the compiler version once benchmarks start failing for that library on that compiler, something in the lines of #92 (benchmark/CMakeLists.txt).

ldionne:
Right; stupid me. I missed this argument (there were many PRs/issues to consider after the weekend).

PR #95, opened 2016-04-04
Load iframes individually instead of batching them

ldionne:
This reduces the time required for loading the iframes when we view the website initially or when we change the compiler/collection. The second commit of this PR is optional; we could also leave the "Loading" dialog but I find it a bit in the way when loading individual iframes, which is quite fast.

I'll wait for your thoughts about this.

### DIFF
--- a/index.html
+++ b/index.html
@@ -178,7 +178,7 @@ compilers:
             <div class="panel-group">
               {% for benchmark in site.documents %}
               <div class="chart panel panel-default" data-collection="{{ benchmark.collection }}" data-compiler="{{ benchmark.compiler }}" style="display: none">
-                <div class="panel-heading" data-toggle="collapse" data-target="#chart{{ forloop.index }}" role="button">
+                <div class="panel-heading" data-toggle="collapse" data-target="#chart{{ forloop.index }}" role="button" onclick="load('#chart{{ forloop.index }}')">
                   <span class="panel-title" >{{ benchmark.url | split: "/" | shift: 3 | pop | join: "/" }}</span>
                 </div>
                 <div id="chart{{ forloop.index }}" class="panel-collapse collapse" style="height: 0px">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -178,7 +178,7 @@
             <div class="panel-group">
               {% for benchmark in site.documents %}
               <div class="chart panel panel-default" data-collection="{{ benchmark.collection }}" data-compiler="{{ benchmark.compiler }}" style="display: none">
-                <div class="panel-heading" data-toggle="collapse" data-target="#chart{{ forloop.index }}" role="button">
+                <div class="panel-heading" data-toggle="collapse" data-target="#chart{{ forloop.index }}" role="button" onclick="load('#chart{{ forloop.index }}')">
```
ldionne: There might be a better way to do this?

brunocodutra: That looks perfect to me.

brunocodutra: An alternative is to add the following to `$(function(){...})` (minus typos)

```
$(".chart").on("click", function(){
    load(this);
})
```

brunocodutra:
Great job, it is much simpler this way. I also *think* it works on Firefox out of the box, but I'd advise some testing, because that is a nasty compiler when it comes to `iframe`s. If you don't have easy access to FF leave it on me, no problem.

> we could also leave the "Loading" dialog but I find it a bit in the way when loading individual iframes, which is quite fast.

I don't like the progress dialog myself, but, stupid me, I couldn't think of an alternative.

index.html
brunocodutra: Bottom line: I like separating HTML from code, but it is up to you really.

ldionne: Ah, this is exactly what I was looking for. It appears to work as-is.

ldionne:
> I also think it works on Firefox out of the box, but I'd advise some testing

It appears to work on Firefox.

brunocodutra:
Also, a bunch of CSS rules should be obsolete now, I think you can safely erase lines 68-86

brunocodutra:
And also the inline `style="height: 0px"` on line 184

index.html
```diff
@@ -65,26 +65,6 @@
         text-align: right;
       }
 
-      .panel-group .panel-collapse.collapse {
```
ldionne: @brunocodutra Can you confirm this is what you meant? I'm completely illiterate in CSS, so I just followed your instructions blindly.

brunocodutra: Looks good.

PR #98, opened 2016-04-05
Add new benchmark type/replace_if

brunocodutra:


brunocodutra:
@ldionne I'm deliberately not implementing benchmarks using Hana, because I'm certain you would do a much better job.

PR #99, opened 2016-04-06
Replace the FILENAME argument by OUTPUT

ldionne:
This is more flexible, since we can specify the directory where we'd like charts to be generated.


brunocodutra:
Looks good.
Once this is merged I'll rebase #98.

ldionne:
Actually, I'll merge yours and then modify this one. That's what I already did locally.

brunocodutra:
>  Actually, I'll merge yours and then modify this one. That's what I already did locally.

Even better :p
Thanks

PR #100, opened 2016-04-06
Split dependencies into separate files

ldionne:


brunocodutra:
Much cleaner, thanks.

Issue #71, opened 2016-03-28
[gh-pages] Provide benchmark data through some public API


brunocodutra:
Following on from #69, we should provide some public API that others could use load benchmark data dynamically into their projects, something in the lines of [GH's API](https://developer.github.com/v3/).

ldionne:
I think just providing the JSON files at some documented URL would be enough. Also, I don't think we can do better with GitHub Pages. After all, it's meant to be a site web publishing service, not a full blown server.

brunocodutra:
> I think just providing the JSON files at some documented URL would be enough.

Forgive my ignorance of web technologies, [I just realized](http://stackoverflow.com/questions/18923328/is-there-a-way-to-enable-cors-on-github-pages) one may dynamically load JSON files hosted on gh-pages using `$.getJSON`. That fits perfectly what I meant by "some public API", I just thought it wouldn't be as easy as simply serving JSON files at some URL.

ldionne:
No problem.

Issue #34, opened 2016-03-16
Allow creating charts to display executable size


ldionne:
Right now, we gather data for the compilation time and the executable size, but we only ever plot the compilation time. It would be nice to allow selecting which aspect is displayed on the chart. I see the following possible designs here:

1. The aspect being plotted is set "statically", i.e. it is set somewhere and the generated `chart.html` displays that and only that aspect. To set the aspect being plotted, we can either set it in the call to `metabench_add_benchmark` by means of `metabench_add_benchmark(name [COMPILATION_TIME|EXECUTABLE_SIZE] ...)`, or we can set it in the `chart.json` file.

2. The aspect being plotted is "dynamic", in the sense that the chart contains all the necessary data and allows switching between aspects using e.g. a tooltip.

I would personally tend towards (1), because it seems more useful to me.

brunocodutra:
I'm ok with displaying only one aspect, but I'd be against controlling it via `metabench_add_benchmark` as CMake should be strictly responsible for configuring the benchmark itself.

ldionne:
Ok, so that's something that should be setup in `chart.json`? And it would default to showing the compilation time?

Note that I was initially in favour of using `metabench_add_benchmark` to control this, but I understand your point. My motivation was strictly convenience; it would be handy to have a way to specify the aspect without having to create a separate `chart.json` file with perhaps only one thing in it (the aspect being plotted).

I think one "problem" might be the fact that I'm thinking about `metabench_add_benchmark` like a `metabench_add_chart`, i.e. in my view the "benchmark" is actually tied to the chart. That might be wrong, though.

brunocodutra:
> Ok, so that's something that should be setup in chart.json? And it would default to showing the compilation time?

Yes, a new property `aspect: <time|size>` should be enough.

> I think one "problem" might be the fact that I'm thinking about metabench_add_benchmark like a metabench_add_chart, i.e. in my view the "benchmark" is actually tied to the chart. That might be wrong, though.

Interesting, I see it rather as `gather_benchmark_data_for([...] DATASETS [...])` whose main objective is to export data as a JSON file that should be easily read by external tools. For convenience it also provides a default tool for visualization which the user may set up via `chart.json`.

> My motivation was strictly convenience; it would be handy to have a way to specify the aspect without having to create a separate chart.json file with perhaps only one thing in it (the aspect being plotted)

[...]

PR #101, opened 2016-04-07
Allow specifying a title in add_benchmark

ldionne:
While this may not be useful for the benchmark suite because of the way http://ldionne.com/metabench is setup, I need this capability for some uses of Metabench in presentations.


brunocodutra:
Just for the record, I'm totally in favour of providing additional features in order to make sure Metabench is useful beyond the benchmark suite.

That said, I think we should broaden the scope of this PR to tackle #60, so basically we have to:

1. Stop generating the aggregate JSON file and, from now on, regard `metabench_add_benchmark` as semantically equivalent to "add chart". 
2. Get rid of `chart.json` and, instead, add new parameters to `metabench_add_benchmark` which would be used for the configuration of the output HTML file (the first of which is `TITLE`).
3. Make the individual JSON files produced by each `dataset` public (and ultimately serve them through `gh-pages` as in #71)

What do you think?

ldionne:
> What do you think?

I think that is the right thing to do.

ldionne:
I think getting rid of `chart.json` and adding new parameters to `metabench_add_benchmark` is a nontrivial task. Let's say we want to allow setting the axis labels. Right now, we're also setting the tick precision because we know the Y axis represents the time. But what if this is not true anymore? I suggest we defer the task of adding more parameters to another PR, and keep `chart.json` for the time being.

brunocodutra:
Actually I think the opposite, we should not give the user too much control over NVD3, because it can break very easily.

Tick precision for example should never be set by the user, since it's Metabench who's in control of what data is generated. IMO it should be set by `metabench_add_benchmark` in accordance to the parameter which controls whether we want a chart for time, binary size or whatever (`ASPECT` perhaps?).

Do you see any need for anything besides `TITLE`, `SUBTITLE`, `XLABEL`, `YLABEL` and `APSECT`? These seem reasonably easy to setup.

ldionne:
I agree with you, and I think the customization points you suggest should suffice (at least for now). Also, if we have `ASPECT`, I don't think we need `YLABEL`, since these are somewhat equivalent.

brunocodutra:
> Also, if we have ASPECT, I don't think we need YLABEL, since these are somewhat equivalent.

Good point, it can certainly be defaulted in accordance to `ASPECT`, however I think we should still provide it in case the user wants to provide labels in a different idiom for example.

ldionne:
Ok; I'll handle your points (1) and (3) above in PR #103. This PR will handle point (2).

ldionne:
This will also address #34.

ldionne:
@brunocodutra By the way, the title is sometimes overlapping the legend, but I don't know how to workaround it:
screen shot 2016-04-08 at 12 20 57

brunocodutra:
> By the way, the title is sometimes overlapping the legend, but I don't know how to workaround it

Well, we could lower the title about `1em`, would that work for you?

ldionne:
I lowered the subtitle too, and it appears to work as a charm.

ldionne:
If you're fine with this (i.e. I did not butcher the JS chart creation too much), I'll merge this.

brunocodutra:
> If you're fine with this (i.e. I did not butcher the JS chart creation too much), I'll merge this.

Looks great

PR #102, opened 2016-04-07
[Travis] Only test one compiler on Ninja

ldionne:
This will significantly speed up the Travis builds without really reducing our coverage, because compilers are likely to behave the same regardless of the CMake generator.


brunocodutra:
Actually I think we should do the opposite, i.e., benchmark on Ninja, because I feel benchmarks run much slower using `make`,  at least locally.

ldionne:
This is probably because Ninja will automatically run many jobs in parallel. While it's nice to know that we support it, we must also avoid this when generating the benchmarks for gh-pages because we want to reduce the load on the CPU as much as possible.

brunocodutra:
> This is probably because Ninja will automatically run many jobs in parallel.

That must be it then. 

> we must also avoid this when generating the benchmarks for gh-pages because we want to reduce the load on the CPU as much as possible

Indeed, lets stick to `make` then.

PR #105, opened 2016-04-08
Import benchmarks from Hana to the benchmark suite

ldionne:
This is based on PR #101.

### DIFF
--- a/benchmark/hetero/count_if/CMakeLists.txt
+++ b/benchmark/hetero/count_if/CMakeLists.txt
@@ -1,0 +1,10 @@
+# Copyright Louis Dionne 2016
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+
+set(datasets)
+add_dataset(datasets dataset.hetero.count_if.fusion.list   fusion.list.cpp.erb   "(0...50).step(5)")
+add_dataset(datasets dataset.hetero.count_if.fusion.vector fusion.vector.cpp.erb "(0...50).step(5)")
+add_dataset(datasets dataset.hetero.count_if.hana.tuple    hana.tuple.cpp.erb    "(0...50).step(5).to_a + (50..400).step(25).to_a")
+
+add_chart(benchmark.hetero.count_if DATASETS ${datasets})
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -50,8 +50,15 @@ endmacro()
 
 add_custom_target(benchmarks)
 macro(add_chart target)
-    metabench_add_chart(${target} ${ARGN} OUTPUT index.html)
-    add_dependencies(benchmarks ${target})
+    cmake_parse_arguments(ARGS "" "" "DATASETS" ${ARGN})
+
+    # If we have no datasets, perhaps they were all disabled in the current
+    # configuration. In this case, we omit the benchmark instead of failing.
+    list(LENGTH ARGS_DATASETS _n_datasets)
+    if (_n_datasets GREATER 0)
+        metabench_add_chart(${target} ${ARGN} OUTPUT index.html)
+        add_dependencies(benchmarks ${target})
+    endif()
 endmacro()
 
 
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -10,11 +10,12 @@ option(METABENCH_METAL      "Benchmark Metal"   ON)
 option(METABENCH_MPL        "Benchmark MPL"     ON)
 
 
-include(benchmark/dependencies/boost)
 include(benchmark/dependencies/brigand)
+include(benchmark/dependencies/fusion)
 include(benchmark/dependencies/hana)
 include(benchmark/dependencies/meta)
 include(benchmark/dependencies/metal)
+include(benchmark/dependencies/mpl)
 include(benchmark/dependencies/std)
 
 
--- a/benchmark/dependencies/fusion.cmake
+++ b/benchmark/dependencies/fusion.cmake
@@ -2,18 +2,22 @@
 # Distributed under the Boost Software License, Version 1.0.
 # (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 
-if (METABENCH_FUSION)
+# Fusion is broken on VS 12.0
+if (METABENCH_FUSION AND NOT (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" AND
+                              CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19"))
     find_package(Boost QUIET)
     if (Boost_FOUND)
         message(STATUS "Local Fusion installation found - Boost ${Boost_VERSION}")
+        function(Fusion_add_dataset dataset)
+            metabench_add_dataset(${dataset} ${ARGN} MEDIAN_OF 3)
+            target_include_directories(${dataset} PUBLIC ${Boost_INCLUDE_DIRS})
+        endfunction()
     else()
         message(STATUS "No local Boost installation found - Fusion will be unavailable.")
     endif()
 endif()
 
-function(Fusion_add_dataset dataset)
-    if (Boost_FOUND AND METABENCH_FUSION)
-        metabench_add_dataset(${dataset} ${ARGN} MEDIAN_OF 3)
-        target_include_directories(${dataset} PUBLIC ${Boost_INCLUDE_DIRS})
-    endif()
-endfunction()
+if (NOT Fusion_add_dataset)
+    function(Fusion_add_dataset)
+    endfunction()
+endif()
--- a/benchmark/dependencies/fusion.cmake
+++ b/benchmark/dependencies/fusion.cmake
@@ -1,0 +1,19 @@
+# Copyright Louis Dionne 2016
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+
+if (METABENCH_FUSION)
+    find_package(Boost QUIET)
+    if (Boost_FOUND)
+        message(STATUS "Local Fusion installation found - Boost ${Boost_VERSION}")
+    else()
+        message(STATUS "No local Boost installation found - Fusion will be unavailable.")
+    endif()
+endif()
+
+function(Fusion_add_dataset dataset)
+    if (Boost_FOUND AND METABENCH_FUSION)
+        metabench_add_dataset(${dataset} ${ARGN} MEDIAN_OF 3)
+        target_include_directories(${dataset} PUBLIC ${Boost_INCLUDE_DIRS})
+    endif()
+endfunction()
--- a/benchmark/dependencies/mpl.cmake
+++ b/benchmark/dependencies/mpl.cmake
@@ -2,21 +2,14 @@
 # Distributed under the Boost Software License, Version 1.0.
 # (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 
-if (METABENCH_FUSION OR METABENCH_MPL)
+if (METABENCH_MPL)
     find_package(Boost QUIET)
     if (Boost_FOUND)
-        message(STATUS "Local Boost installation found - version ${Boost_VERSION}")
+        message(STATUS "Local MPL installation found - Boost ${Boost_VERSION}")
     else()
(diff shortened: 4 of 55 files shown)


ldionne:
This is still a WIP:
- Some benchmarks are still missing. I'll add more some other time; it's really boring to copy/paste and slightly modify all these files.
- Fusion/MPL benchmarks use C++11/14 features. This is so because of several reasons: first, it was like that in Hana. Second, using C++11/14 features is sometimes necessary to make the benchmarks consistent between libraries. For example, if a benchmark for `count_if` uses `hana::integral_constant` with Hana but uses `mpl::integral_c` with MPL, there's effectively a difference in the benchmark methodology for both. Is this okay? Maybe, but we have to be aware of it. Right now, what I do in these new benchmarks is use `std::integral_constant` all the time, but that requires C++11 even for Fusion/MPL.

Also, almost all benchmarks are imperfect. For example, any benchmark where we create a sequence of length `n` and then perform an operation on it for a varying `n` is actually benchmarking two things: the cost of creating the sequence __and__ the cost of performing the algorithm. For type-only computations this is not too bad, because the cost of creating the sequence is (should be) small. But for heterogeneous computations, creating the heterogeneous sequence can actually be more costly than running the algorithm on it.

I think a solution might be to do the following:

```c++
auto xs1 = hana::make_tuple(x1);
```

brunocodutra:
holy cow

brunocodutra:
>  For example, if a benchmark for count_if uses hana::integral_constant with Hana but uses mpl::integral_c with MPL, there's effectively a difference in the benchmark methodology for both. Is this okay?

Well, regarding this matter, I think it makes more sense to benchmark a library using exclusively the features it provides than porting newer features that technically are not supported, afterall I think the idea is more to provide an insight in the overall performance of a library from the perspective of the user than grind bits to compare specific implementation details.

ldionne:
> I think the idea is more to provide an insight in the overall performance of a library from the perspective of the user than grind bits to compare specific implementation details.

I tend to agree with this. However, for benchmarking stuff like Fusion algorithms, we can't possibly be idiomatic in our usage of the library, since otherwise we wouldn't perform anything (some algorithms return views).

brunocodutra:
>> I think the idea is more to provide an insight in the overall
performance of a library from the perspective of the user than grind bits
to compare specific implementation details.
>
> I tend to agree with this. However, for benchmarking stuff like Fusion
algorithms, we can't possibly be idiomatic in our usage of the library,
since otherwise we wouldn't perform anything (some algorithms return views).

Indeed. How does C++11/14 help enforcing evaluation? I'm not familiar with
fusion, so I'm really clueless.

BTW, it seems that Fusion officially supports C++11 variadic templates, so
we don't need to define that macro prior to including it if we are to stick
to C++11.

Anyways, I like the idea of making it work with C++98 just in case someone
would like to test MPL and Fusion for some old compiler, but that wouldn't
be me and I understand it is a lot of work porting

ldionne:
> Indeed. How does C++11/14 help enforcing evaluation? I'm not familiar with fusion, so I'm really clueless.

It doesn't, sorry. I was unclear, but my comment was related to the idiomatic usage of libraries, not C++14. What I meant is that more generally, we won't be able to benchmark every single library in the most idiomatic way to use it, because micro-benchmarks would otherwise be impossible for some libraries (e.g. Fusion). Instead, we have to force the evaluation using `fusion::as_list(fusion::some_algo(fusion::make_list(...)))`. This is suboptimal, but if we don't do it then we won't have anything to benchmark because `fusion::some_algo` returns a lazy view.

> BTW, it seems that Fusion officially supports C++11 variadic templates, so we don't need to define that macro prior to including it if we are to stick to C++11.

Good point. Also, it is very important that we use the variadic versions of the containers, because otherwise the compile-times we'll report for Fusion will be much worse than they should be. I think the variadic implementations are used whenever the compiler supports them, but I'll check.


> Anyways, I like the idea of making it work with C++98 just in case someone would like to test MPL and Fusion for some old compiler, but that wouldn't be me and I understand it is a lot of work porting

There is some amount of work required for porting, but what I'm even more worried about is the fact that using C++98 for some benchmarks and C++14 for others could lead us into biasing the benchmarks in subtle ways.

ldionne:
@edouarda @ericniebler Since you guys represent like 30% of the potential users of Metabench's benchmark suite, are you interested in seeing the results of the benchmark suite on older compilers? If not, I think that removes a noticeable incentive for us to support C++98 whenever possible.

ericniebler:
Support C++98? Which of our libraries even compiles with C++98?

ldionne:
None of ours, but MPL and Fusion do.

ldionne:
Basically, the question is: Are you interested in knowing how MPL and Fusion performs on old compilers?

brunocodutra:
> what I'm even more worried about is the fact that using C++98 for some
benchmarks and C++14 for others could lead us into biasing the benchmarks
in subtle ways.

Perhaps I wasn't clear, what I mean to say is that it would be nice if the
benchmark suite ran on whatever compiler the user might try. Technically
the only thing that prevents it from working on, say, GCC 3 is the fact
Fusion and MPL benchmarks explicitly make use of C++11 features. Other than
that CMake should successfully disable every library but MPL and Fusion
because of lack of C++11 support (well technically we hard test compiler
versions now, but you know what I mean).

I understand there might be no  actual use for this, but who knows, perhaps
one day I just feel like porting everything down to C++98 just for the fun
of it. Or not.

brunocodutra:
> Also, almost all benchmarks are imperfect. For example, any benchmark where we create a sequence of length n and then perform an operation on it for a varying n is actually benchmarking two things: the cost of creating the sequence and the cost of performing the algorithm. For type-only computations this is not too bad, because the cost of creating the sequence is (should be) small. But for heterogeneous computations, creating the heterogeneous sequence can actually be more costly than running the algorithm on it.

Can't we just benchmark a reference file which simply creates the sequence for the given `n` and subtract the result from the actual benchmark?

ldionne:
> Perhaps I wasn't clear, what I mean to say is that it would be nice if the benchmark suite ran on whatever compiler the user might try. [...]

I understand.


> I understand there might be no actual use for this, but who knows, perhaps one day I just feel like porting everything down to C++98 just for the fun of it. Or not.

I have no _personnal_ interest in Metabench supporting C++98 compilers. I'm not sure what you suggest is a sufficient rationale for porting Fusion/MPL benchmarks to C++98. What I'd suggest is that if you ever find yourself wanting to implement your suggestion above, then you can just as well port the benchmarks to C++98 at that moment. But doing the work right now, presuming we might need it at some point in the future, I'm not excited about that.

> Can't we just benchmark a reference file which simply creates the sequence for the given n and subtract the result from the actual benchmark?

That might work. I'm concerned that launching a separate compilation would introduce a large amount of variability in the results, but perhaps not. We'd have to experiment. If this works, then it's definitely a better idea than my original proposal.

ldionne:
Regarding C++98/11/14: What I'd suggest is that we set a hard limit of basic C++11 support for the benchmarks (even for Fusion/MPL). Basically, anything that works for the set of compilers we're interested in (i.e. the benchmark suite) should be good enough. I would see anything more as being too much work for the added value.

edouarda:
@ldionne I wouldn't spend a lot of time trying to get the bench work on old compilers.

brunocodutra:
>  What I'd suggest is that if you ever find yourself wanting to implement your suggestion above, then you can just as well port the benchmarks to C++98 at that moment.

That's what I'm trying to say. 

> Regarding C++98/11/14: What I'd suggest is that we set a hard limit of basic C++11 support for the benchmarks (even for Fusion/MPL). 

I agree and suggest we set the limit based on the compilers we actually test on Travis and Appveyor.

> That might work. I'm concerned that launching a separate compilation would introduce a large amount of variability in the results, but perhaps not. 

It certainly will, but perhaps the median filter could help.

ldionne:
Ok, perfect. Then no additional work will be done to make the Fusion/MPL benchmarks I added in this PR C++98-compliant.

ldionne:
@brunocodutra Any idea why the Appveyor build is failing? It does not seem to be related to any particular library, but to some strange CMake issue. See [this](https://ci.appveyor.com/project/ldionne/metabench/build/1.0.387/job/qaokby0x9yhe68qh#L2116) and [this](https://ci.appveyor.com/project/ldionne/metabench/build/1.0.389/job/lel8xn0qkl9275k1#L1955).

brunocodutra:
> Any idea why the Appveyor build is failing?

I think there are two separate issues going on here. 

First of all, MSVC 12 consistently fails the same Fusion benchmarks most certainly due to its poor compliance to C++11. Maybe we should consider dropping support for it before we start bending over backwards to please a broken compiler. 

That aside, it looks like another issue occasionally causes all benchmarks of a library to fail at the first iteration. I'm not sure what could be the culprit, but I'd guess there's something going wrong with the external projects?

Perhaps something prevents the compressed package from being extracted and CMake is unable to notice that something went wrong. What if we add a custom [test step](https://cmake.org/cmake/help/v3.0/module/ExternalProject.html) to assert everything we need is there? That wouldn't solve the problem, but should make it easier to debug at least.

@ericniebler This got me thinking, is MSVC 14 able to handle Meta?

ericniebler:
> This got me thinking, is MSVC 14 able to handle Meta?

I don't believe so.

brunocodutra:
>> This got me thinking, is MSVC 14 able to handle Meta?
>
> I don't believe so.

Right, I was just afraid I might have wrongly got to that conclusion after I saw Meta benchmarks failing on Appveyor.

brunocodutra:
@ldionne Forget what I said, just found out MSVC 14 has been updated on Appveyor this weekend and it seems [they broke it real good](https://ci.appveyor.com/project/brunocodutra/metal/build/404/job/dn2dy8phmy6qsug4) this time.

Feel free to disable Metal on MSVC till I've figured it out

benchmark/hetero/count_if/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# Copyright Louis Dionne 2016
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+
+set(datasets)
+add_dataset(datasets dataset.hetero.count_if.fusion.list   fusion.list.cpp.erb   "(0...50).step(5)")
+add_dataset(datasets dataset.hetero.count_if.fusion.vector fusion.vector.cpp.erb "(0...50).step(5)")
+add_dataset(datasets dataset.hetero.count_if.hana.tuple    hana.tuple.cpp.erb    "(0...50).step(5).to_a + (50..400).step(25).to_a")
```
brunocodutra: I think we should uniformize range across benchmarks, so that charts look somewhat similar.
I'd also be in favor of using a piecewise defined range in order to both reduce the time required to run the benchmark suite and also to increase the range, perhaps something like

`(0...99).step(10).to_a + (100...1000).step(50).to_a`

brunocodutra: ... at least among benchmarks of the same category, i.e. pure type and heterogeneous.

brunocodutra:
Since brunocodutra/metal#26 Metal should compile on MSVC 14 again, however it doesn't compile on GCC 4.8 anymore.

@ldionne Could you adjust the benchmark suite accordingly?

benchmark/hetero/count_if/CMakeLists.txt
ldionne: I agree this is a good idea; let's do this in a separate PR though. For now all I want is to port Hana's benchmarks to Metabench, even if they are not perfect.

brunocodutra: Sounds good.

brunocodutra:
Looks like appveyor takes so long cloning Boost that it is stalling, so how about we use [preinstalled versions](https://www.appveyor.com/docs/installed-software#languages-libraries-frameworks)? 

Contrary to Travis, which is quite limited by LTS Ubuntu, the Appveyor team is quite free to keep their worker images pretty up to date, so preinstalled Boost versions shouldn't lag that much behind.

ldionne:
Agreed, it makes me crazy. See #107.

brunocodutra:
Anything further preventing this PR from being merged?

ldionne:
It's still missing some benchmarks, but I'll merge it so you can start adding Metal to what's there. I'll submit another PR with more benchmarks eventually.

brunocodutra:
> It's still missing some benchmarks, but I'll merge it so you can start adding Metal to what's there. I'll submit another PR with more benchmarks eventually.

Thanks, that's the idea

Issue #106, opened 2016-04-14
Uniformize the dataset ranges


ldionne:
In [this comment](https://github.com/ldionne/metabench/pull/105#discussion_r59610455) at #105, Bruno says

> I think we should uniformize range across benchmarks, so that charts look somewhat similar. I'd also be in favor of using a piecewise defined range in order to both reduce the time required to run the benchmark suite and also to increase the range, perhaps something like
>
> `(0...99).step(10).to_a + (100...1000).step(50).to_a`

I think this is a good idea.

ldionne:
Fixed by #113.

PR #108, opened 2016-04-15
Allow mitigating the base cost of benchmarks

ldionne:
Right now, benchmarks look like

```c++
#include <boost/hana/count_if.hpp>
#include <boost/hana/tuple.hpp>

#include <type_traits>
namespace hana = boost::hana;


struct is_even {
    template <typename N>
    constexpr bool operator()(N const&) const {
        return N::value % 2 == 0;
    }
};

int main() {
    auto ints = hana::make_tuple(
        <%= (1..n).map { |i| "std::integral_constant<int, #{i}>{}" }.join(', ') %>
    );
    auto result = hana::count_if(ints, is_even{});
    (void)result;
}
```

The problem with this is that we're benchmarking both the cost of creating the sequence and that of the `count_if` algorithm. This is misleading, since the cost of creating the sequence is almost always much larger than the cost of performing the algorithm. This causes almost all the benchmarks to look like they have a gentle linear performance, while this really is the creation of the sequence. To mitigate this, it was suggested that we benchmark the cost of creating the sequence itself, and subtract it from the cost of the whole.

The goal of this PR is to explore whether this seems useful, and whether subtracting the base cost is a reliable way of achieving our goal. Right now, this is only a WIP. For a dataset with a source file named `xxx.cpp.erb`, if you provide a  `xxx.cpp.erb.base` file, that file will be compiled like the normal `.cpp.erb` file and its compile-time will be subtracted from it. If we go this way, I have something else in mind for the implementation, but for now we should just figure out whether this is desirable.

@brunocodutra Your thoughts on this would be appreciated.


brunocodutra:
If I'm not mistaken, the base value should be shared by every benchmark, that is, once we've measured the time to create the sequences for all `n` of interested, we can basically subtract that very same result from every other benchmark accordingly.

Now, since the "base benchmark" (wouldn't bias sound more accurante BTW?) will be affecting every other benchmark, I believe it would be wise to enable a considerably large median filter for it, say 5 or even 7. 

The time requirements might be prohibitive, however, so it might be necessary to reduce the density of samples, which IMO is currently very high, so I don't think that would be a no-go.

brunocodutra:
Here's an idea:

How about we create a further parameter `BASE_DATASET` for `metabench_add_chart`? We would need to bascially load its JSON file and subtract it from every other dataset, sample by sample, before generating the HTML file. This way a base benchmark could easily be setup just like any other benchmark

BTW, couldn't we just use the `make` benchmark we already have as a base for all other hetero benchmarks?

ldionne:
Our minds seem to converge; this is the "something else I have in mind" I hinted about before. However, I have the following reserves:

1. This means that whatever base dataset we're using, it's going to have to be generated for all the values of `n` that we use across all the benchmarks.

2. Different benchmarks create slightly different sequences. For example, the `count_if` benchmark uses a `hana::make_tuple(std::integral_constant<int, 1>{}...)`, while the `transform` benchmark uses a `hana::make_tuple(x<1>{}...)`. Some benchmarks might also `#include` slightly different files. This could cause the base adjustment to be (perhaps subtly) wrong. Do we care about this? The other solution is to explicitly state the code that should be used as a base for each dataset.


Something I've thought about if we want to specify the base for each dataset: we could use a macro to delimit the "base" part of the benchmark:

```c++
#include <boost/hana/count_if.hpp>
#include <boost/hana/tuple.hpp>

#include <type_traits>
namespace hana = boost::hana;


struct is_even {
    template <typename N>
    constexpr bool operator()(N const&) const {
        return N::value % 2 == 0;
    }
};

int main() {
    auto ints = hana::make_tuple(
        <%= (1..n).map { |i| "std::integral_constant<int, #{i}>{}" }.join(', ') %>
    );
    // this would be defined only when benchmarking the "base" part of a file
#if !defined(METABENCH_BENCHMARK_BASE) 
    auto result = hana::count_if(ints, is_even{});
    (void)result;
#endif
}
```

This way, we would at least not need to write too much duplicate code. In summary, the two solutions I see are

1. Reuse the same dataset as a base for all benchmarks that want it

2. Provide a way to exclude base code on a per-dataset basis

(1) will be faster when running the whole benchmark suite. (2) is simpler from a Metabench user perspective, and I think it would be more useful for someone doing simple throwaway benchmarks. (2) is also more precise, although we could achieve the same precision by creating a base dataset tailored to every single dataset (but that would be more verbose).

brunocodutra:
> Our minds seem to converge; this is the "something else I have in mind" I hinted about before. 

Awesome; in fact I did suspect it were.

> This means that whatever base dataset we're using, it's going to have to be generated for all the values of n that we use across all the benchmarks.

I think we just need to carefully craft ranges to make sure they match each other.
This relates to #106. 

> Different benchmarks create slightly different sequences. For example, the count_if benchmark uses a hana::make_tuple(std::integral_constant<int, 1>{}...), while the transform benchmark uses a hana::make_tuple(x<1>{}...). 

I don't think it would be too hard to make most benchmarks use the same sequences. Take your example, for instance, we could just as well [define a metafunction for `x<_>`](https://github.com/ldionne/metabench/blob/master/benchmark/type/replace_if/hana.types.cpp.erb). Sure that won't apply to evey case, take `at` for instance which should use a sequence of the same size for all `n`, but perhaps in this case we want a constant OFFSET instead of a BASE_DATASET.

> Some benchmarks might also #include slightly different files. This could cause the base adjustment to be (perhaps subtly) wrong. Do we care about this? The other solution is to explicitly state the code that should be used as a base for each dataset.

You are right, each DATASET has its own base, I mean, it makes no sense to subtract the same base value from all datasets that compose a benchmark. Think MPL which preprocess a whole bunch of nasty macros, if we subtracted it from brigand for example we will most probably see negative values.

> In summary, the two solutions I see are
> 1. Reuse the same dataset as a base for all benchmarks that want it

IMO this won't work for the reasons above.

> 2. Provide a way to exclude base code on a per-dataset basis

I think we can achieve it by adding a new parameter BASE or BIAS to `metabench_add_dataset`, which should point to a `cpp.erb` template. Naturally several datasets could use the same template as a BASE and it must be guaranteed that a given template will only be compiled once, i.e., during compilation of the first dataset that uses it. The way we setup `metabench_ada_dataset` I believe this would be easy to guarantee by relying on CMake and the makefile generator directly.

ldionne:
Just before we dive too deep into implementation strategies, I think it would be good to confirm that we want the feature. Here is the chart generated for `count_if` with the median of 3 trick. screen shot 2016-04-15 at 15 52 33

Blue is the total compile-time without adjustment, green is the cost of creating the sequence that we'll call `count_if` on, and orange is the compile-time of calling `count_if` _and_ subtracting the cost of creating the sequence. I think this feature will make the benchmarks much more reliable.

brunocodutra:
>  I think this feature will make the benchmarks much more reliable.

Looks promising indeed.

ldionne:
This is superseded by #109.

PR #109, opened 2016-04-15
Add support for microbenchmarks

ldionne:
This PR is a WIP presenting an alternative implementation to #108. I prefer this implementation because it is simpler to use and seems much more useful. The only downside is that it might cause the benchmark suite to take longer to execute, but that probably isn't a deal breaker.


brunocodutra:
If we go this way, which I think is fine, I'd rather make it the only behaviour, that is, always require that any benchmark explicitly tell what exactly to benchmark.

ldionne:
Hmm, but sometimes you really want to benchmark the whole file. The following is still left to do:

- [ ] Determine whether we make it the default behaviour
- [ ] Settle on the name of the option, if we keep one
- [ ] Properly document the `MICROBENCHMARK` option (if we keep it)
- [ ] Settle on the name of the macro to define. Right now, I use `METABENCH_BASE`, but it's not great
- [ ] Add unit tests
- [ ] Delimit what is to be benchmarked in all the benchmarks of the suite

brunocodutra:
> Hmm, but sometimes you really want to benchmark the whole file. 

Yes, and that takes just 2 extra lines.

```
#if defined(METABENCH)
...
#endif
```

On the other hand we avoid maintaining and documenting a two-in-one cmake function.

ldionne:
Ok, I guess that could make sense. But how would it work? Do we define a macro (which?) when we want to benchmark the base, or the contrary?

ldionne:
And what macro name would we use?

brunocodutra:
> Do we define a macro (which?) when we want to benchmark the base, or the contrary?

The most obvious I think would be to define something like `METABENCH` to denote when we are running for real, i.e., the negated logic of your original proposal.

Alternatively we could use some kind of ERB marker, but I wouldn't know which or how it would work or even if that makes sense at all, but I trust you'd know the answer to these questions.

ldionne:
I decided against the ERB marker, because that would have made it impossible to put `#ifdef METABENCH` somewhere inside a header that's included in the `.cpp.erb` file for local throwaway benchmarks.

ldionne:
Any idea why the Appveyor build [is failing](https://ci.appveyor.com/project/ldionne/metabench/build/1.0.423/job/1nsi0l52uamdhute#L690)?

brunocodutra:
> Any idea why the Appveyor build is failing?

Could it be running out of RAM perhaps?

ldionne:
I would definitely be surprised if the error `ninja: error: opening build log: Permission denied` was caused by a lack of RAM. I'll do some experiments tomorrow.

brunocodutra:
> I would definitely be surprised if the error ninja: error: opening build log: Permission denied was caused by a lack of RAM.

Indeed it seems unrelated, but maybe that could be a catch-all-error-message.

ldionne:
Just linking this PR to the discussion [here](https://github.com/ldionne/metabench/commit/b4d043d5d006a616f913674df4c9791002b9ff08#commitcomment-17163264).

ldionne:
Ok, so from their reply in the issue above, it seems like we're on our own. We know what the problem is likely to be, but we don't have a clear fix except that we shouldn't be doing what we're doing. Unfortunately, I don't see any other way of doing what we're doing than doing it the way we are (if that makes sense).

If we create a separate CMake project and add the benchmarks to it, we could have a separate build directory for Ninja. But then the benchmark/dataset targets wouldn't be part of the current CMake project, which isn't useful. Otherwise, we can also drop support for Ninja, but that would be __lame__. So I'm not really sure what to do here. By the way, it seems like we already have this problem with our current design, it just didn't manifest itself.

Any ideas?

brunocodutra:
Well, I wouldn't start spending time on a work around for this issue just yet.

I think our best shot for now would be to try NMake once again so we can go ahead till we have a decent fix for Ninja.

BTW, is there a way to avoid logging in Ninja?

brunocodutra:
And just so you know, I'll be kindda inaccessible for a week or so.

ldionne:
NMake [doesn't work](https://ci.appveyor.com/project/ldionne/metabench/build/1.0.485/job/5bgi0dedvv4vh50m).

> And just so you know, I'll be kindda inaccessible for a week or so.

Sure, no problem.

ldionne:
> BTW, is there a way to avoid logging in Ninja?

I don't think so. `ninja --help` doesn't say so, at least. Plus, I think the `build.log` is kinda important to Ninja's functioning, it doesn't seem like something that would be optional.

ldionne:
So this has been merged, but I'd like you to take a look at the live benchmarks and tell me what you think. There are some negative timings, and the timings for small sequences don't seem to provide us with much insight. I think this PR is a good idea, but we might be missing something to make the results more precise?

brunocodutra:
The only inconvenience I see is the fact we now entirely ignore the preprocessor, so MPL benchmarks in particular are little representative of the end user experience.

ldionne:
Right. I'm not 100% satisfied with the current state of things, but I think that's the best we can do without actually modifying the compiler.

PR #113, opened 2016-04-17
Uniformize ranges

brunocodutra:
The total number of samples were decreased roughly fivefold, while at the same time `MEDIAN_OF 3` was enabled for all libraries. The result are smother benchmarks and a considerable speed up when running the entire benchmark suite.

This PR addresses #106.

type
hetero


ldionne:
Nice!

PR #123, opened 2016-05-25
Add IDE folders in CMake.

mkurdej:
Folders are working on MSVC only for the moment, they are used to regroup projects.


ldionne:
Is Metabench even working with the MSVC generator? I think it only works with Makefile generators, no?

brunocodutra:
> I think it only works with Makefile generators, no?

Yes, because we rely on [RULE_LAUNCH_LINK](https://cmake.org/cmake/help/v3.0/prop_gbl/RULE_LAUNCH_LINK.html), which is ignored by non-makefile generators.

> Is Metabench even working with the MSVC generator?

NMake is a makefile generator for MSVC, so in theory it should work, but I don't think it does, because it sucks.

Now I don't know how that relates to `USE_FOLDERS`, in fact I'm not sure what it does either, is it so we could call `make benchmarks.type` to only build the targets within that directory?

ldionne:
@mkurdej Can you please explain what your PR does? Unfortunately, it is my first time encountering the `FOLDER` CMake property, and since I don't use Visual Studio, I have no idea what it does concretely.

mkurdej:
Folders are only used to group projects in a MSVC solution, nothing less, nothing more.
I'm not sure but it may work as well in other IDEs (XCode?).

Actually, I was trying to make metabench work with MSVC non-NMake generator and found it easier to do it with folders instead of flat project hierarchy.

ldionne:
Ok, so I tried your patch with Xcode, and it does change the organization of the project. But it does not do so in a helpful manner, because the targets are still flat:
screen shot 2016-05-26 at 12 34 07

I assume this is better on MSVC? Also, I fail to build any target with Xcode, which seems to support the claim that Metabench will only work with Makefile generators. Did you manage to build the benchmarks with MSVC in Visual Studio?

mkurdej:
Yes, it looks better on MSVC.
I haven't managed to build teh benchmarks with VS, but I hope that I'll manage to do it soon.

metabench-ide-folders

ldionne:
Ok. Well, I have two comments:

1. I'm not sure about the added value of this PR. Indeed, the tests and examples are basically run by Travis CI and that's it. Except for experimentation, this repository should probably not be cloned and used as-is. So improving the layout of the project on MSVC seems like a marginal gain. The only thing that I think might be worthwhile to a wider audience is to place charts created with `metabench_add_chart` in a common folder, but that raises the question of how should that folder be named.

2. Until we manage to make Metabench work on VS, there is basically no added value. So the first step should be to make it work on VS, and then we could consider making it _prettier_ on VS.

Does that seem reasonable?

brunocodutra:
> Except for experimentation, this repository should probably not be cloned and used as-is

That's mostly how I use metabench locally TBH, so, even though I don't use VS myself, I can understand the value of this PR for the poor souls who do.

That said...

> Until we manage to make Metabench work on VS, there is basically no added value.

I second that opinion. 

@mkurdej since you mentioned you'd be looking into making metabench work on VS, I propose you extend the scope of this PR so as to effectively add support for VS (be it through NMake or otherwise).

ldionne:
@brunocodutra I'm a bit surprised, but OK. As long as we don't clutter the files too much with `set_properties` calls, if you say that would be useful to you, then OK. Is it possible to centralize the setting of folders or to make it less pervasive?

Of course, I think support for VS should come before though.

brunocodutra:
> if you say that would be useful to you, then OK

That won't be useful to me in particular, but it seems to me that this would be a very simple addition that could prove useful for VS users.

That doesn't mean however that I fully agree with the current implementation. It would be much better to add a further optional parameter `FOLDER` to `metabench_add_dataset`, so that users who integrate `metabench.cmake` into their projects could really benefit from it. 

Does that make sense?

ldionne:
Actually, the dataset targets are not public right now, so I don't think we should allow putting them into folders. However, the chart targets are public, and it might make sense to put them into folders. But I don't think that adding an option to `metabench_add_chart` is the right way to go, as it is just as easy to use a separate `set_target_properties`.

metabench.cmake
```diff
@@ -300,6 +300,7 @@ function(metabench_add_chart target)
     else()
         add_custom_target(${target} DEPENDS "${ARGS_OUTPUT}")
     endif()
+    set_target_properties(${target} PROPERTIES FOLDER "charts")
```
brunocodutra: this should be set from outside `metabench_add_chart`

brunocodutra:
> Actually, the dataset targets are not public right now

What you mean by public?

> I don't think that adding an option to `metabench_add_chart` is the right way to go, as it is just as easy to use a separate `set_target_properties`

You got a point there, the feature doesn't really need any support from our side. 
In this case I'm OK with it as long as there are no modifications to `metabench.cmake` and as long as it works on VS of course.

ldionne:
> What you mean by public?

Part of Metabench's API. Actually, dataset __targets__ do not even exist. What exists is chart targets, and JSON files containing the generated datasets. But the JSON files are not targets, they're only CMake entities that you can be dependent upon. So the only thing that we _might_ want to put into folders, AFAIU, are the chart targets.

ldionne:
Oh, and Metabench's unit tests.

ldionne:
Are there any plans to expand this to support MSVC? If not, I'll close for lack of clear added value.

mkurdej:
For the moment, I don't have time to work on this, but I'll certainly work on support for MSVC in the (not so distant) future. Go ahead and close this PR, I'll open a new one (or reopen this one) when the time comes.

ldionne:
Alright then. Please let us know when/if you start working on MSVC support and need any help. I think it would be very valuable for us to support it.

PR #128, opened 2016-07-31
update pure type benchmarks

brunocodutra:
This PR makes sure pure type benchmarks are up to date with the recent breaking changes in Metal.


ldionne:
Thanks a lot for the cleanup.

Issue #124, opened 2016-05-29
Metabench is measuring the wrong thing


ericniebler:
Very few people are manipulating type lists of > a few dozen elements. I find the current tests to be largely meaningless for real world metaprograms. It would be more interesting to run the tests up to, say, 100, but repeatedly and with lots of different types to force the number of instantiations up. 

In addition, benchmarks are meaningless if the total CPU time is less than about 5s. So, for each N in [0,100] each test should keep adding unique instantiations of the algorithm being measured until the CPU time is high enough and then divide by the number of instantiations to get an accurate idea of the performance profile of that algorithm at that N.

That way, we can really start the work of optimizing our libs for the real world.

brunocodutra:
> So, for each N in [0,100] each test should keep adding unique specializations of the algorithm being measured until the CPU time is high enough and then divide by the number of specializations to get an accurate idea of the performance profile at that N.

So you propose that, for a given `N`, instead of instantiating the algorithm under test for a single list of size `N`, we instead instantiate it for `K/N` lists of the same size `N` but containing different types (so as to avoid memoization)?

ericniebler:
For each N, instantiate the algorithm with K different lists to avoid memoization. Increase K until CPU time > 5s. Then divide the CPU time by K to get an accurate idea of the cost of each instantiation at that N.

brunocodutra:
> For each N, instantiate the algorithm with K different lists to avoid memoization. Increase K until CPU time > 5s. Then divide the CPU time by K to get an accurate idea of the cost of each instantiation at that N.

Right. I see two problems with this approach as is:

2. `5s` for every `N` for every algorithm for every library would make it impossible to update benchmark results using Travis.
1. Even if we agree on some `X < 5s`, how would we got about *increasing K until CPU time > X*? I mean, Metabench works roughly by generating static source files for each `N` and then measuring the time taken to compile those source files. Sure we could put that in a loop and keep increasing `K` and regenerating the source file until the time taken to compile it gets high enough, but that too would be prohibitively expensive time-wise, even for very small values of `X`.

I think the best we could do would be to define some static `K` (per algorithm) and, for every `N`, generate a source file that instantiates the given algorithm for `K/N` lists of size `N`. Then we measure how long it takes to compile that source file and divide the result by `K/N`. This way we would have roughly the same *resolution* with respect to time for every `N`.

ldionne:
[...]

PR #129, opened 2016-08-01
Make sure MPL benchmarks are representative

brunocodutra:
Contrary to heterogeneous tuples, constructing pure type lists is basically free, so we can really afford to benchmark the preprocessor for pure type libraries. This is particularly important in order for MPL benchmarks to be representative of the end user experience, while Brigand, Meta and Metal are mostly unaffected by these changes. Hana is out of scope here since it is not pure type library, but it shouldn't matter anyways since I don't believe it relies on the preprocessor too much.

### DIFF
--- a/benchmark/type/at/brigand.list.cpp.erb
+++ b/benchmark/type/at/brigand.list.cpp.erb
@@ -2,6 +2,7 @@
 // Distributed under the Boost Software License, Version 1.0.
 // (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 
+#if defined(METABENCH)
 #include <brigand/sequences/at.hpp>
 #include <brigand/sequences/list.hpp>
 #include <brigand/types/integer.hpp>
@@ -12,7 +13,6 @@ using list = brigand::list<
     <%= (1..n).map { |i| "x<#{i}>" }.join(', ') %>
 >;
 
-#if defined(METABENCH)
 using result = brigand::at<list, brigand::int32_t<<%= n - 1 %>>>;
 #endif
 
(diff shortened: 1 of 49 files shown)


ldionne:
The reason for not benchmarking the includes and the creation of the type lists is that these are meant to be microbenchmarks of the algorithms. For example, say we create a sequence once and include the MPL headers once, but run the measured algorithm 200 times and want to know how that's going to behave. Then, clearly, measuring the include time and sequence creation is wrong, and we must only measure the compilation time of the algorithm itself.

That being said, if we want to provide benchmarks that are a better indicator of the end user experience (which I think is highly desirable, and is related to #124), then we should devise macro benchmarks that solve some problem from end-to-end with a given library, and then measure that as a whole.

brunocodutra:
This is a bit of a conundrum indeed. 
The time taken to preprocess header files is only paid for once, so strictly speaking it should not be considered as part of the benchmark of a single algorithm, but then how do we go about the fact MPL benchmarks we provide are basically a lie? The viewer might be led to believe MPL does actually do pretty well compared to modern libraries, when in fact just including MPL headers takes an order of magnitude longer than algorithms themselves.

ldionne:
Sure, I understand and agree with you. I think the way we go about this is by providing actual user-experience benchmarks. We find a classical TMP problem and solve it from end to end with a given library. Then, we benchmark the whole thing, including preprocessing and so on. On the website, we have two sections
1. micro benchmarks, which is relevant to us implementing TMP libraries
2. macro benchmarks, which is relevant for users wanting to know the user experience of each library

brunocodutra:
> I think the way we go about this is by providing actual user-experience benchmarks. We find a classical TMP problem and solve it from end to end with a given library. Then, we benchmark the whole thing, including preprocessing and so on

Oh now I see what you mean, sounds like a good idea.

PR #131, opened 2016-08-01
Avoid erasing previous benchmarks if something goes wrong

brunocodutra:
It's better to show outdated benchmarks than no benchmarks at all

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,10 +113,10 @@ script:
   # triggers the nightly build.
   - |
     if [[ "${METABENCH_NIGHTLY_BUILD}" == "true" && "${CMAKE_GENERATOR}" == "Unix Makefiles" ]]; then
-      cmake --build build --target benchmarks
+      cmake --build build --target benchmarks &&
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -113,10 +113,10 @@ script:
   # triggers the nightly build.
   - |
     if [[ "${METABENCH_NIGHTLY_BUILD}" == "true" && "${CMAKE_GENERATOR}" == "Unix Makefiles" ]]; then
-      cmake --build build --target benchmarks
+      cmake --build build --target benchmarks &&
```
ldionne: Shouldn't the script already exit as soon as a command returns with a non zero status? This is strange, do you know why we're removing everything when it fails?

brunocodutra: > Shouldn't the script already exit as soon as a command returns with a non zero status?

Yes, **BUT** technically nothing fails, the `if` command still returns success. The thing is Travis doesn't see what happens withing block statements.

ldionne: Ah, yes, obviously. Thanks for catching.

PR #132, opened 2016-08-03
[Travis] Add support for Clang 3.9

ldionne:
[Clang 3.9 is currently being tagged](http://llvm.1065342.n5.nabble.com/llvm-dev-3-9-Release-Release-Candidate-1-has-been-tagged-tp99158.html). Once it is officially released, we can start testing on it using this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ addons:
       - clang-3.6
       - clang-3.7
       - clang-3.8
+      - clang-3.9
       - g++-4.8
       - g++-4.9
       - g++-5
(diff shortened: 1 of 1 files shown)


brunocodutra:
I was able to run unit tests for Metal [on Clang 3.9 and even on 4.0](https://github.com/brunocodutra/metal/blob/develop/.travis.yml) by requiring `sudo`. 

Granted it [should be slower than the container based infrastructure](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/), but I wonder if it would be really *prohibitively* slow for metabench, I mean, we could always split pure type and hetero into their own jobs. Have you ever tested `sudo: required` for metabench?

PR #133, opened 2016-08-09
updated find_if to use new brigand lambdas

porkybrain:


porkybrain:
brigand changing to its new lambda style breaks some of the benchmarks because they were using the nested apply template which is no longer supported. This should fix that (provided the checks pass ;))

benchmark/type/find_if/brigand.list.cpp.erb
```diff
@@ -7,17 +7,15 @@
...
+template<typename N>
+using is_last = brigand::bool_<N::type::value == <%= n %>>;
 
 using list = brigand::list<
     <%= (1..n).map { |i| "brigand::int32_t<#{i}>" }.join(', ') %>
 >;
 
 #if defined(METABENCH)
-using result = brigand::find<list, is_last>;
+using result = brigand::find<list, brigand::bind<is_last, brigand::_1>>;
```
ldionne: Do we need to include something such as `brigand/bind.hpp`?

porkybrain: the algorithms themselves (replace_if, find, etc.) should have all the needed headers.

ldionne: Alright, thanks.

PR #139, opened 2016-08-19
Add benchmarks for all

brunocodutra:
Meta is disabled because it currently fails this benchmark with the following error on Clang 4.0

> error: pack expansion used as argument for non-pack parameter of alias template
>        `using invoke = typename F::template invoke<Args...>;`

@ericniebler did I do anything wrong?

**Note:** This is based on #137


ericniebler:
No, you did nothing wrong. I've opened an issue here: https://github.com/ericniebler/meta/issues/44

PR #144, opened 2016-08-20
Smooth benchmark results

brunocodutra:
This PR reflects [this comment](https://github.com/ldionne/metabench/issues/69#issuecomment-241218287) on #69

### DIFF
--- a/metabench.cmake
+++ b/metabench.cmake
@@ -208,6 +208,7 @@ endfunction()
 #                     [TITLE title]
 #                     [SUBTITLE subtitle]
 #                     [XLABEL label] [YLABEL label]
+#                     [INTERPOLATE]
 #                     DATASETS dataset1 [dataset2 [dataset3 [...]]]
 #                     [OUTPUT path/to/file])
 #
@@ -251,6 +252,10 @@ endfunction()
 #   [YLABEL label]:
 #       The label to use for the Y axis.
 #
+#   [INTERPOLATE]:
+#       If `INTERPOLATE` is provided, the datasets will be interpolated into
+#       seemingly continuous smooth lines.
+#
 #   DATASETS dataset1 [dataset2 [dataset3 [...]]]:
 #       A list of datasets from which the benchmark is generated.
 #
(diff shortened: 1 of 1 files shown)


brunocodutra:
Thinking about it, it should be made optional whether any kind of interpolation should be applied.

brunocodutra:
This adds an option `SMOOTH` to `metabench_add_chart`, which enables D3 *bundle* interpolation for all datasets in a chart.

This option is set for all charts generated by the benchmark module.

brunocodutra:
I realized this new property would be better named `INTERPOLATE` and it should be clear that it does not affect the underlying data. I'll amend it later today.

Apart from that, is there anything preventing this from being merged? Even if we decide not to set this option for the benchmarks we display online, I think this is an useful option to have.

ldionne:
No, no outstanding issue, but I'm not sure we want to enable this for the benchmark suite. I believe that having the raw data points would be better than simply trends if we could increase the precision of benchmarks for small sequences.

brunocodutra:
>  I believe that having the raw data points would be better than simply trends if we could increase the precision of benchmarks for small sequences.

Better yet if we can have both. Notice this interpolation here only afects how data is *visually displayed* in the chart in an effort to make it easier for a human to make sense of it, however we still have the original raw data available through the JSON files for machine processing if so desired (we still have to publish them on the `gh-pages` btw). That is to say I really don't think there's any value added to displaying raw data for humans to see, because we just see noise there.

ldionne:
I know we're not modifying the raw data in any way. What I meant is that it's valuable to see the exact point on the chart if we have a lot of precision. Say I want to pick the library that has the best behavior for `reverse` for sequences under 30 elements. If I only have trends, I don't have a clear way to make a decision for small sequences.

brunocodutra:
> What I meant is that it's valuable to see the exact point on the chart if we have a lot of precision.

Indeed, that's exactly how I use metabench locally to fine tune metal against the other libraries, but that only works because locally I can control the noise, i.e., by not touching my laptop while benchmarks are running. I just don't think we can get to that level of precision using a virtual machine in the cloud which does not guarantee we are the single clients of the physical processor, so I think it would be better to advertise those results as a simple indication of the trend, so users don't lose their faith in them when they see negative values or otherwise nonsense timings that decrease with the increase of the data size.

ldionne:
I think we can fight this by benchmarking the same algorithm on several different sequences instead, thus reducing the noise even for small sequences.

brunocodutra:
I suppose you refer to #148? All right, sice we agree the option to interpolate results is useful regardless, I'll update this PR disabling the interpolation for the benchmark suite so that this can be merged and we postpone the decision regarding the consequences of enabling it for on the `gh-pages`.  Sounds reasonable?

ldionne:
Sounds perfect!

brunocodutra:
Done

PR #146, opened 2016-08-23
Speed up travis builds

### LINKED ISSUE #145: Travis time limit blown up
After the recent inclusion of new algorithms some builds have been [blowing up Travis time limit](https://travis-ci.org/ldionne/metabench/jobs/154068674).

brunocodutra:
Addresses #145

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,8 +90,8 @@ install:
     fi
 
   # Install Boost headers
-  - BOOST_URL="http://github.com/boostorg/boost.git"
-  - travis_retry git clone --depth 1 --recursive --quiet ${BOOST_URL}
+  - BOOST_URL="https://sourceforge.net/projects/boost/files/latest/download"
+  - mkdir boost && travis_retry wget --no-check-certificate --quiet -O - ${BOOST_URL} | tar --strip-components=1 -xj -C boost
   - (cd boost && ./bootstrap.sh && ./b2 headers)
 
   # Install and use a more recent Ruby


brunocodutra:
Maybe we should implement other changes on top of downloading boost as a compressed package, because we might find ourselves fighting this beast once again in the near future once we improve benchmarks coverage.

ldionne:
While this is going to help for sure, I don't think it will address #145 since it only saves a few minutes. Cloning Boost was just taking ~144s, which is not much in the grand scheme of things.

brunocodutra:
Oh I didn't mean this first commit fixed things once and for all, hence my
followup comment, I just meant to say it was part of the solution.

Maybe I didn't choose words well :p

ldionne:
Oh, alright. Anyway thanks, that will make things slightly better. I guess another possible improvement might be to split the generation of GCC benchmarks in multiple jobs. Not sure how difficult that would be, though.

brunocodutra:
Well I'm wary of these kinds of ad-hoc workarounds at a first glace because
they often only serve to delay the problem and make things harder to fix
then.

Personally I suspect the root cause of this issue lies on the fact
something is suboptimal in our setup so that we are somehow wasting
valuable resources for meager gains. Personally I think the median filter
is to blame here and I was originally planing to investigate an alternative
which could deliver similar results but that wouldn't cost that much
timewise. Anyways, I'm open to discussing this.

In the mean time I'd just lower the size of datasets, specially for some of
the hetero benchmarks which are quite expensive, to get builds running till
we settle on a definitive fix.

brunocodutra:
Come to think about it, adapting the travis script to build hetero and pure
type benchmarks on separate buids should be an even simpler and less
intrusive workaround for the time being.

ldionne:
I'm running the benchmarks locally with `time` to see which ones are really expensive. If it's mostly the heterogeneous benchmarks, then perhaps reducing them would be a good idea. We should also try to see whether there's any noticeable gain from taking the median of 3. If not, we should make a single measurement since that'll cut the time by 3 without much effort.

PR #147, opened 2016-08-24
Make it possible to build pure type and hetero benchmarks individually

brunocodutra:
New intermediary custom targets `benchmarks.type` and `benchmarks.hetero` have been created to make it possible to better fine tune what is built.

This PR also reduces clutter and improves maintainability by avoiding abusing regular expressions to parse dot separated strings, since we can have arguments passed individually from the beginning.

Changes are extensive, but the bulk of it is a simple find and replace to split dot separated strings as individual arguments.


ldionne:
While I think the ability to build hetero and pure-type benchmarks is very useful, I find the proliferation of so many arguments to be somewhat confusing. Wouldn't it be possible to simply say something like this?

```cmake
if (${target} MATCHES "benchmark.hetero")
    add_dependencies(benchmark.hetero ${target})
endif()
```

brunocodutra:
Well to be honest this refactoring was more a planed feature than a by
product of the title of this PR.

See I was kindda annoyed by the redundancy of stating
`add_dataset(dataset.something...` every time and somewhat worried by the
fact we relied on regex too much. It is very easy to match more than
intended and I have been bitten by this before. Finally I don't see the
advantage of concatenating arguments as a long string and parsing it back
to pieces, if we can have cmake handle multiple arguments nicely for us.

Think of it as a bottom up approach where we feed in individual pieces and
assemble the information internally as opposed to the top down approach
where you feed in everything fit together and then pull things apart to get
the information you need. From the point of view of software design it is
easier to handle simple pieces and make sure they are assembled correctly
internally, than expect the user to do so externally and risk breaking
things while looking for the internal pieces based on assumptions of how it
should have been assembled. A great example of this were the benchmarks
at_key-deep and at_key-wide, which couldn't be called at_key.deep and
at_key.wide as one would naturally expect, because that little dot there
would break regular expressions further deep the call stack.

benchmark/CMakeLists.txt
```diff
@@ -19,29 +19,29 @@ include(benchmark/dependencies/mpl)
 include(benchmark/dependencies/std)
 
 
-macro(add_dataset datasets dataset path_to_template range)
-    string(REGEX REPLACE "[^.]+[.][^.]+[.][^.]+[.](.*)" "\\1" name ${dataset})
-    set(args ${dataset} ${path_to_template} ${range} NAME ${name} OUTPUT ${name}.json ${ARGN})
+macro(add_dataset datasets type case lib datatype path_to_template range)
+    set(name "${lib}.${datatype}")
+    set(dataset "datasets.${type}.${case}.${name}")
```
ldionne: Please use singular `dataset.` instead of `datasets.`

ldionne: Also, `type` is not a great name since it's very general. Perhaps `category` would be better?

benchmark/CMakeLists.txt
```diff
@@ -50,15 +50,23 @@ macro(add_dataset datasets dataset path_to_template range)
 endmacro()
 
 add_custom_target(benchmarks)
-macro(add_chart target)
+macro(add_chart type case)
```
ldionne: `category` instead of `type`?

benchmark/CMakeLists.txt
```diff
@@ -50,15 +50,23 @@ macro(add_dataset datasets dataset path_to_template range)
...
     cmake_parse_arguments(ARGS "" "" "DATASETS" ${ARGN})
 
+    set(type_target "benchmarks.${type}")
```
ldionne: This should be singular `benchmark.`, not `benchmarks.`.

benchmark/CMakeLists.txt
```diff
@@ -50,15 +50,23 @@ macro(add_dataset datasets dataset path_to_template range)
...
+
     # If we have no datasets, perhaps they were all disabled in the current
     # configuration. In this case, we omit the benchmark instead of failing.
     list(LENGTH ARGS_DATASETS _n_datasets)
     if (_n_datasets GREATER 0)
-        metabench_add_chart(${target} ${ARGN} OUTPUT index.html)
-        add_dependencies(benchmarks ${target})
+        metabench_add_chart(${case_target} ${ARGN} OUTPUT index.html)
+        add_dependencies(${type_target} ${case_target})
+        add_dependencies(benchmarks ${type_target})
```
ldionne: What happens when we call `add_dependencies` multiple times on the same targets? Here, we'll call `add_dependencies(benchmarks benchmarks.hetero)` and `add_dependencies(benchmarks benchmarks.type)` multiple times.

benchmark/CMakeLists.txt
ldionne: Ah, actually we probably want `benchmarks.hetero` (because it's many), but `benchmark.hetero.fold_left` (because it's only one). Or is it more consistent to be plural/singular all the time?

brunocodutra: My reasoning is that it makes sense to think of target names in this context in analogy to filesystem paths, i.e. the longer the prefix, the more specific it is, up to the point where it denotes a single entity. Notice that `benchmarks.hetero.at` is really just the same thing as `documents/research/some_report.txt`.

benchmark/CMakeLists.txt
brunocodutra: Nothing special AFAICT

benchmark/CMakeLists.txt
ldionne: OK

ldionne:
Merged in c44e9bb3b36f328302ac50026ab95e4b9d0f704a, thanks. I was unsure about the split, but turns out I like it after playing around locally. I did make all the target names singular, though (see c72311be45c3f9761c4ff2fdb6c2a466e90be425).

Issue #148, opened 2016-08-24
Consider adding high-precision microbenchmarks for small sequences


ldionne:
Forking #124 to discuss high precision microbenchmarks for small inputs more precisely. The idea is that we'd like to have a precise view of the behavior of algorithms on small sequences, because this is how they are mostly (but not only) used in the wild. I see two different approaches:

1. For each sequence length `N`, generate `K` different sequences of length `N` and call the algorithm once on each sequence. By having a sufficiently large `K`, the total compilation time is increased and the relative error is reduced.
2. Same as (1), but then divide the result by `K` to find an approximation of the absolute time taken for a single algorithm. I have reservations with this approach, because the compilation time as we increase `K` (the number of small sequences) is not necessarily linear.

brunocodutra:
Would `K` be constant for all `N` for a given algorithm? If so, I'd be ok with the first option. If not I think we would have to stick with the second one, because otherwise charts would be non-intuitive to the viewer.

ldionne:
> Would `K` be constant for all `N` for a given algorithm?

We can choose to go either way. I agree that decreasing `K` as `N` increases would help keep the build times short and having a large `K` for large `N`s would not be useful. However, I'm unclear on how we would specify which `K` to associate to which `N`.

brunocodutra:
> I'm unclear on how we would specify which `K` to associate to which `N`

I believe choosing `K` so that `K*N = <some constant>` should be good enough.

> We can choose to go either way. 

So let us take a step at a time. Since the first option is basically a special case of the second one, why don't we implement `1.` and then, if necessary, refine it into `2.`?

ldionne:
> So let us take a step at a time. Since the first option is basically a special case of the second one, why don't we implement `1.` and then, if necessary, refine it into `2.`?

Sounds good. That raises (at least) one question: do we want to tweak the existing benchmarks so that they have higher precision for smaller sequences, or do we want to create a new set of benchmarks for "small sequences"? The first option would obviously be better if that can be done, but that might end up taking too much Travis time, depending on how smart we can be.

brunocodutra:
> do we want to tweak the existing benchmarks so that they have higher precision for smaller sequences, or do we want to create a new set of benchmarks for "small sequences"?

[...]

PR #149, opened 2016-09-05
Made metabench use the system + user time for his measurements instea…

CvRXX:
…d of just the raw time. This means that the time will be less influenced by other processes runnig on the system. Idea for this commit came from @chieltbest


brunocodutra:
Sounds awsome, does it improve results noticeably?

ldionne:
I just generated the benchmarks with this patch on my laptop, and the results actually appear to have more variability.

ldionne:
Actually, it could be that real time is measured with more precision, which means that _total_ time ends up giving more steps in the charts. I do agree that, in theory, total time is most likely what we should be measuring, but in practice it does not seem to give the best results on my laptop. I'll launch the benchmarks on Travis and push them temporarily to some dummy branch so we can look at them.

CvRXX:
Looking forward to the results!

CvRXX:
user + system time is less accurate so you will get more variability in small tests but there will be less noise from other processes in bigger tests. Maybe make a switch for it?

ldionne:
Actually, it appears to significantly improve the reliability of results. See this, where the left hand side is the benchmarks generated with this patch, and the right hand side is without the patch: 
metabench

brunocodutra:
can we afford to disable `MEDIAN_OF` and solve #145 for now?

ldionne:
> can we afford to disable MEDIAN_OF and solve #145 for now?

Checking: https://travis-ci.org/ldionne/metabench/builds/158330425

brunocodutra:
Am I screwing something up or are results exactly the same on branch `gh-pages` and `gh-pages-nomedian`?

ldionne:
IDK what went wrong; running again: https://travis-ci.org/ldionne/metabench/builds/158625013

ldionne:
Ok; I just compared the results and it appears that we're not significantly worse without the median filter. In fact, the variability of our data is huge in all cases for small sequence sizes, which sucks. I think #148 would add a lot of value to the suite. Either way, I'll disable the median filter for the benchmark suite and resolve #145 for now.

PR #150, opened 2016-10-11
Improve the resolution of benchmark measurements

brunocodutra:
This is a work in progress towards #148.

The idea is to impose a common bias to all benchmarks in order to increase compilation times thus allowing for better measurements.


brunocodutra:
@ldionne before I proceed with the rather tedious task of manually updating every single benchmark, do you think you could set up a travis run to push results for the updated fold_left benchmark to a dummy branch, so that we could assess whether `k` in this order of magnitude is enough to improve results?

ldionne:
The problem is that I believe the Travis job will not be able to push to GitHub since it's started by a pull request, and so it does not have my GitHub API key. Let me look.

ldionne:
I've made you a collaborator, which means that you get push access to the repository. Please do not
1. push to master without my consent
2. leak the `GITHUB_TOKEN` in the `.travis.yml`; I'm trusting you on this one

I've also created two branches: `gh-pages-high_resolution` and `brunocodutra/high_resolution`. You can rebase your changes on top of `brunocodutra/high_resolution` and push to that branch as much as you want. I have it setup so that `gh-pages-high_resolution` gets updated whenever you push to `brunocodutra/high_resolution`, so you can experiment. Once we're done, we can delete both branches and you can update this PR. Does that seem reasonable?

ldionne:
Oh, and thank you a bunch for tackling this.

brunocodutra:
> I've made you a collaborator, which means that you get push access to the repository

Thanks I appreciate it.

> Once we're done, we can delete both branches and you can update this PR. Does that seem reasonable?

Sounds good to me.

brunocodutra:
Right now unit tests alone are taking over 15 minutes to complete on travis (clang 3.8) and that is likely to grow past 20 minutes once we are done with this PR. Is there any strong reason against disabling unit tests on nightly builds?

ldionne:
It's nice to be warned by the nightlies if we break something in Metabench. What's the issue with nightlies taking a long time to run? If you want, you can disable the UTs on your branch for testing purposes.

brunocodutra:
> It's nice to be warned by the nightlies if we break something in
Metabench.

Sure, but we are already warned by the regular builds if we  break
metabench directly. What could possibly break nightlies are indirect
changes in the libraries themselves, but that will cause benchmarks
themselves to fail anyways. In other words, unit tests are redundant on
nightlies.

> What's the issue with nightlies taking a long time to run?

Jobs will time out. They already take ~45 minutes to complete with high
resolution applied to the pure variant only.

ldionne:
You're right, I'll disable the unit tests during nightly builds.

brunocodutra:
Great, thanks.

benchmark/hetero/at/CMakeLists.txt
```diff
@@ -3,10 +3,10 @@
 # (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 
 set(datasets)
-add_dataset(datasets hetero at fusion list        fusion.list.cpp.erb      "[1] + (10..50).step(10).to_a")
-add_dataset(datasets hetero at fusion vector      fusion.vector.cpp.erb    "[1] + (10..50).step(10).to_a")
-add_dataset(datasets hetero at hana   basic_tuple hana.basic_tuple.cpp.erb "[1] + (10..50).step(10).to_a + (100..500).step(100).to_a")
-add_dataset(datasets hetero at hana   tuple       hana.tuple.cpp.erb       "[1] + (10..50).step(10).to_a + (100..500).step(100).to_a")
-add_dataset(datasets hetero at std    tuple       std.tuple.cpp.erb        "[1] + (10..50).step(10).to_a + [100]")
+add_dataset(datasets hetero at fusion list        fusion.list.cpp.erb      "(0..50).step(10).to_a"                            ENV "{k: 3, p: 100}")
```
ldionne: Why don't we look at 100 for `fusion::list`?

benchmark/hetero/at/fusion.vector.cpp.erb
```diff
@@ -2,24 +2,22 @@
 // Distributed under the Boost Software License, Version 1.0.
 // (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
 
-<% if n > 10 %>
-    #define FUSION_MAX_VECTOR_SIZE <%= ((n + 9) / 10) * 10 %>
```
ldionne: How come is this not required anymore? After all, we go until 100 elements for `fusion::vector`, and that should blow up the default limit, no?

ldionne:
Also, with these changes, I'm not sure whether the axis on the charts is still relevant. The y axis is till time, but the x axis is not simply the number of elements anymore. I see mainly the following solutions:

1. Do some processing on the measured times to try to get the time for a specific number of elements. For example, let's say we take `hetero.make`, we could divide the obtained time by `k` to get the actual time for creating a single sequence. However, this does not always make sense, for example with `at` where both the size of the sequence, the number of sequences and the number of indices we access in each sequence changes.
2. Instead of trying to display accurate x-axis values, we could instead go for something bold and just say __small input__, __medium input__ and __large input__, or perhaps something finer grained.
3. Do nothing. I feel like this is not great, since that gives us the wrong impression in terms of absolute times.

benchmark/hetero/at/fusion.list.cpp.erb
```diff
@@ -2,24 +2,22 @@
...
+    /*constexpr*/ auto <%= "xs#{k}" %> = fusion::make_list(
+        <%= ((k*n+1)..(k*n+n)).map { |n| "x<#{n}>()" }.join(', ') %>
     );
 
-#if defined(METABENCH)
-    fusion::at_c<<%= n - 1 %>>(xs);
-#endif
+    <% (env[:p]*n/100).times do |p| %>
+    #if defined(METABENCH)
+        fusion::at_c<<%= p*100/env[:p] %>>(<%= "xs#{k}" %>);
```
brunocodutra: Not exactly, that actually means `p%` elements equally distributed in the range `[0, n-1]` are retrieved, so, for instance, if `n = 60` and `p = 10%`, then indices will be `[0, 10, 20, 30, 40, 50]`.

benchmark/hetero/at/CMakeLists.txt
brunocodutra: You mean `n = 100`? Because otherwise it would take a disproportional amount of time.
I tried selecting `n` so that results are all at a similar scale thus preventing a single dataset from dominating the chart.

benchmark/hetero/at/fusion.vector.cpp.erb
brunocodutra: It was [never required](http://www.boost.org/doc/libs/1_62_0/libs/fusion/doc/html/fusion/container/vector.html):

> For C++11 compilers, the variadic function interface has no upper bound.

brunocodutra:
> Also, with these changes, I'm not sure whether the axis on the charts is still relevant. The y axis is till time, but the x axis is not simply the number of elements anymore. I see mainly the following solutions:

> 1. Do some processing on the measured times to try to get the time for a specific number of elements. For example, let's say we take `hetero.make`, we could divide the obtained time by `k` to get the actual time for creating a single sequence. However, this does not always make sense, for example with `at` where both the size of the sequence, the number of sequences and the number of indices we access in each sequence changes.

TBH I think this is valid even for `at`, because the only thing we vary is still `n`, since the proportion of elements accessed with respect to the size of the sequences is fixed. In other words it benchmarks the retrival of `p*n/100` elements from a sequence, where `n` is the only variable as usual.

brunocodutra:
> Also, with these changes, I'm not sure whether the axis on the charts is still relevant. 

fixed by bc90ca4

benchmark/hetero/at/CMakeLists.txt
ldionne: Ok, I see.

benchmark/hetero/at/fusion.list.cpp.erb
ldionne: Ah, now I see this.

metabench.cmake
```diff
@@ -426,10 +434,10 @@ file(WRITE ${METABENCH_RB_PATH}
 "      IO.write(cpp_file, code)                                                                           \n"
 "      results = repetitions.times.map { build(target) }                                                  \n"
 "      datum = {}                                                                                         \n"
-"      datum['compilation_times'] = results.map { |r| r['compilation_time'] }                             \n"
-"      datum['link_times']        = results.map { |r| r['link_time'] }                                    \n"
-"      datum['executable_size']   = results.map { |r| r['executable_size'] }.first                        \n"
-"      datum['execution_times']   = results.map { |r| r['execution_time'] }                               \n"
+"      datum['compilation_times'] = results.map { |r| (scale)*r['compilation_time'] }                     \n"
```
ldionne: Instead, would it make sense to pass `datum['scale']` to the JavaScript and do the scaling over there?

brunocodutra: I thought about that at first, but in this particular case I saw no benefit in doing so

ldionne: Ok, let's keep it this way for now.

benchmark/CMakeLists.txt
```diff
@@ -18,11 +18,19 @@ include(benchmark/dependencies/metal)
 include(benchmark/dependencies/mpl)
 include(benchmark/dependencies/std)
 
-macro(add_dataset datasets category case lib datatype path_to_template range)
+macro(add_dataset datasets category case lib datatype path_to_template k range)
+    set(env "{k: ${k}")
```
brunocodutra: > I'm not a big fan of specifying additional ENV values as a CMake list

Me neither, but I am even less a fan of the obvious alternative to be redundant and specify both `k` and `scale` and *regexping* a string to match key value pairs between curly braces is simply insane. 

It might not be pretty but that is the CMake syntax for pairs.

ldionne:
Is it OK if I remove the `gh-pages-high_resolution` and `brunocodutra/high_resolution` branches?

brunocodutra:
> Is it OK if I remove the gh-pages-high_resolution and brunocodutra/high_resolution branches?

Sure, please do so.

Issue #119, opened 2016-04-30
Consider allowing microbenchmarks to be disabled


ldionne:
I encountered a case where the microbenchmark thing actually caused the gathered data to be nonsense. The problematic benchmark is in [this](https://gist.github.com/ldionne/feddaf8f0530cb36bab84026215df9e4) Gist. In that benchmark, I use a slightly modified version of Metabench which does not subtract the base time of the compilation, because otherwise the curves are nonsense. Either there's something I don't understand about that benchmark, or there's something fishy with our implementation of microbenchmarks and it might be wise to allow switching it off to simply measure a whole file.

ldionne:
It just occurred to me that instead of removing the baseline in the Ruby script, we could provide both measurements to the Javascript and we could then do the calculation there instead. Each chart could then provide the ability to show either with or without the baseline, or something like that.

In a similar vein, we could perhaps forward _all_ data points to the Javascript and apply any filters there. This would give us the flexibility to show error bars and stuff like that. I don't think we need this for now, but this is something to keep in mind.

brunocodutra:
Great idea!

ldionne:
This is obsoleted by #155, which makes it possible to turn microbenchmarks on/off per graph.

PR #152, opened 2016-10-16
Provide all the data to the JavaScript and calculate the median there

ldionne:
As discussed in #119, providing all the data to the JavaScript will make it easier for us to customize the charts to show/hide the base cost and doing other clever tricks.


brunocodutra:
Looks great, thanks for looking into this.

PR #153, opened 2016-10-26
User controls

brunocodutra:
This PR is the first step towards a richer user interface with better tools to analyse benchmark data. Here we add a simple control button to charts that allows the user to turn B-spline interpolation of data samples on and off.

Some followup ideas:
* let the user select among [a handful of interpolation methods](https://github.com/d3/d3-3.x-api-reference/blob/master/SVG-Shapes.md#line_interpolate);
* let the user turn filtering on and off;
* let the user select the filtering algorithm.

image
image


ldionne:
Awesome! I'm not sure I would go as far as giving the control of the algorithm to the user, however, since that might be slightly overwhelming. I think another valuable option would be to control whether the base of the benchmark is subtracted.

brunocodutra:
> I think another valuable option would be to control whether the base of the benchmark is subtracted.

Excellent idea!

PR #155, opened 2016-10-29
Extend user controls to allow turning debiasing on and off

brunocodutra:
Here debiasing stands for subtracting the baseline from benchmark results.

debiased

biased


ldionne:
I think it would be more intuitive if the option said "subtract baseline" instead of "debias". What do you think?

brunocodutra:
> What do you think?

I think you are much better than me at picking names.

ldionne:
Aweeeesome, thanks!

PR #160, opened 2016-11-22
Change the interpolation mode from 'basis' to 'cardinal'

chieltbest:
Doing so will provide a accurate match to the data, while preserving
the desired smoothing behavior.

A better option might be to provide options for all three
interpolation modes ([none](https://github.com/d3/d3-shape/blob/master/README.md#curveLinear), [cardinal](https://github.com/d3/d3-shape/blob/master/README.md#curveCardinal), [B-spline](https://github.com/d3/d3-shape/blob/master/README.md#curveBasis)), as the B-spline
interpolation is still desirable under certain circumstances such as
data with a high amount of noise.

For a reference of all the interpolation modes see:
https://github.com/d3/d3-shape/blob/master/README.md#curves

### DIFF
--- a/metabench.cmake
+++ b/metabench.cmake
@@ -608,7 +608,7 @@ file(WRITE ${CHART_HTML_ERB_PATH}
 "             });                                                                                                           \n"
 "      }                                                                                                                    \n"
 "                                                                                                                           \n"
-"      chart.interpolate('basis').useInteractiveGuideline(true);                                                            \n"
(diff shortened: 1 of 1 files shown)


ldionne:
Apologies for the slow turnaround time. I like this, @brunocodutra what do you think?

brunocodutra:
> Apologies for the slow turnaround time. I like this, @brunocodutra what
do you think?

Spikes don't seem to be such a critical problem anymore, so I think we can
afford this indeed.

Issue #115, opened 2016-04-21
Ninja fails to open the build log on Windows


ldionne:
When we use a high value for the `MEDIAN_OF` argument, Ninja errors out on Windows. An example can be seen [here](https://ci.appveyor.com/project/ldionne/metabench/build/1.0.487/job/495gi9vfqdto7cam#L69).

This bug was already discussed in PR #109, ninja-build/ninja#1132, and in [these comments](https://github.com/ldionne/metabench/commit/b4d043d5d006a616f913674df4c9791002b9ff08#commitcomment-17163264). I'm creating a separate issue because the issue is already present (even before #109), it just didn't trigger, and I want to handle it separately so we can move on with #109.

ldionne:
Ok, so I went ahead and disabled the Appveyor build temporarily in a3ecf8d9b2d4d7a19cc51736db109fcbacfda5ed. This will allow us to proceed with #109, and we can then try to find a solution for this issue.

PR #163, opened 2017-01-14
[appveyor] Test against Boost 1.63.0

ldionne:


ldionne:
Support on Windows is broken anyway until #115 is fixed. Closing this.

PR #168, opened 2017-03-11
added library Kvasir

odinthenerd:


odinthenerd:
older clang failing, funny clang 3.7 with microsoft codegen works, guess I need to install clang 3.6. Is there a way to exclude tests on old clang for now ?

ldionne:
Try looking at what we do for Hana on old GCCs

odinthenerd:
looks like the porblem occurs with our wrappers around meta continuations. here is a short example of the problem https://godbolt.org/g/l2jkT8 I think we can solve this without too much trouble.

brunocodutra:
Reposting from cpplang for future reference:

> working around the issue with the fixed type is easy, just make sure the nested alias template is dependent somehow, like this: https://godbolt.org/g/8dQoAc

odinthenerd:
clang 3.5, 3.6, 3.7 will be slower because I force a lazy instantiation to get around the alias flaw but should work now

ldionne:
If you're ready to merge, can you please squash into a sensible set of commits? Just edit your branch locally and then push-force. That will update the PR; no need to create a new PR.

brunocodutra:
and update the copyright notices on top of each file

odinthenerd:
OK ready to go, I'm always a bit timid when it comes to --force ;)

PR #171, opened 2017-03-17
[benchmarks] Fix find_if benchmarks

### LINKED ISSUE #170: Benchmarks for find_if are broken
As pointed out by @pdimov at pdimov/mp11#3

> the benchmarks for find_if don't look correct, they look for `n` with the intent to find the last element, but the list contains `n*k+1..n*k+n`, so they don't.

ldionne:
Fixes #170.

Once this passes, I'll leave one day so everyone can check whether they are happy with the way the benchmarks look for their library (to make sure nobody is disadvantaged), and then I'll merge if nobody complains.


brunocodutra:
Thanks for looking into this so quickly, looks good to me.

PR #172, opened 2017-03-20
[brigand] Fix include path.

mkurdej:
Fix include path for Brigand as mentioned in https://github.com/edouarda/brigand/issues/243.


ldionne:
Thanks!

PR #174, opened 2017-03-21
Add mp11

brunocodutra:


brunocodutra:
@odinthenerd is that all you had so far?

benchmark/dependencies/mp11.cmake
```diff
@@ -0,0 +1,37 @@
+# Copyright Odin Holmes 2017
+# Distributed under the Boost Software License, Version 1.0.
+# (See accompanying file LICENSE.md or copy at http://boost.org/LICENSE_1_0.txt)
+
+if (METABENCH_MP11
+    AND NOT (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND
+             CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
```
pdimov: 4.7 and 4.8 should work too.

benchmark/dependencies/mp11.cmake
```diff
@@ -0,0 +1,37 @@
...
+    AND NOT (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND
+             CMAKE_CXX_COMPILER_VERSION VERSION_LESS "3.6")
```
pdimov: Here 3.5 works, 3.4 should as well.

brunocodutra: ok

pdimov:
Fails due to missing `<boost/config.hpp>`. :-(

brunocodutra:
Oh good catch, I'll fix that too

brunocodutra:
@ldionne this should be ready to merge if CI passes
@pdimov please shout out if you see anything odd

pdimov:
Seems insert/erase break the g++ builds. I just changed `mp_take` to be less slow though, so they should work better now. If not, let's limit them to something like 200-300.

brunocodutra:
>  so they should work better now
they did, tests passed, thanks

PR #176, opened 2017-03-23
Add a few more pure type benchmarks

brunocodutra:


brunocodutra:
@odinthenerd @pdimov please double check

ldionne:
You guys are on fire! I'm pretty sure we're going to blow up Travis limits soon; we should provision for that.

PR #177, opened 2017-03-24
added make and changed at to use lookup on kvasir

odinthenerd:


odinthenerd:
finally going to see the results of lookup, @brunocodutra sorry for the headache there was a bug breaking everything under 16 elements before, fixed now.

PR #179, opened 2017-03-28
Avoid timeouts on Travis by splitting benchmarks into subparts

### LINKED ISSUE #178: We are blowing up Travis time limits
Lately, we've added a few new metaprogramming libraries and a few new benchmarks. This causes us to blow up the Travis time limits. We should figure out something. Options I see right now:

1. Distribute the work across more jobs (e.g. ~~a job for heterogeneous stuff, and a job for pure-type~~ __that's already what we are doing__).
2. Reduce the sampling size (probably not a good idea)
3. Do not update all libraries all the time (also probably not a good idea)

ldionne:
Will fix #178 when fully working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,30 +39,49 @@ addons:
 
 env:
   global:
+    # Number of parts in which we split the generation of benchmarks to avoid
+    # timeouts on Travis. Keep this in sync with the `PART=X` variables below.
+    - PARTS=3
+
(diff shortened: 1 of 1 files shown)


brunocodutra:
I think this PR broke something, newly pushed benchmarks have no title

ldionne:
Yeah, you're right. Pretty hard to do things right on Travis without any local testing. I'm looking into it.

PR #183, opened 2017-06-18
Metal related adjustments

brunocodutra:
* Disable Metal on GCC < 5 to prevent timing out Travis (see brunocodutra/metal#72)
* Increase the limit for `cartesian_product` to 400 (recently optimized - see local benchmarks below)

__Clang__
image

__GCC__
image


ldionne:
How long does the changes to cartesian_product add to the weekly build?

brunocodutra:
Not much I think, since it takes the same amount of time now for 500 that it took for 200 previously

PR #185, opened 2017-06-24
[Travis] Add Clang 4.0 and GCC 7

ldionne:


brunocodutra:
Out of curiosity, why not running everything on Trusty?

ldionne:
Most of the apt-sources are for Precise, so I wasn't sure they would "just work" on Trusty.

brunocodutra:
Fair enough, one never knows we it comes to Travis

PR #186, opened 2017-07-05
Collect peak memory usage during compilation

### LINKED ISSUE #17: Allow measuring compiler RAM usage
Once things are settled down better, it would be nice to add the ability to measure the RAM usage of the compiler in addition to the compile-times.

ldionne:
This fixes #17 and provides a valuable source of information.

I think the next step would be to always forward all the information to the charts and allow switching between different aspects at the Javascript level instead of hardcoding that in CMake.


ldionne:
@brunocodutra Obviously don't spend any time reviewing until I can fix the Travis build. I'll let you know when I'm ready for you to have a look.

brunocodutra:
No hurry, I won't be able to take a look at it before the end of next week
anyways

PR #195, opened 2018-07-24
added preliminary boost.tmp support

odinthenerd:
this contains a few hetero functions and about half of the type functions. I expect to add the rest at some point in the future in a separate PR


odinthenerd:
hmm looks like I'm crashing pre 4.0 clang, I'll make a workaround in boost.tmp and rerun

odinthenerd:
looks like the last break was a travis bug?

ldionne:
Is this good to go once the Travis is passing?

PR #202, opened 2020-03-01
add CXX_STANDARD and INCLUDE_DIRS parameter to function metabench_add…

aliefhooghe:
add CXX_STANDARD and INCLUDE_DIRS parameter to function metabench_add_dataset


ldionne:
Is there a reason why you don't simply set those properties on the `${target}` itself after calling `metabench_add_dataset`?

It seems like it would be more flexible than adding arguments to `metabench_add_dataset` itself, since there's potentially a lot of properties that we might want to set on it.

aliefhooghe:
I just didn't thought it would work ! Thanks for the tip, it is much simpler like that.

